PR #2, opened 2018-06-16
Fix compilation, update to Kotlin 1.2.50

jansorg:
- Move BytePacketBuild() into common modules to avoid a JS name clash 
- Update to latest versions


qwwdfsad:
We are completely reworking current implementation, so the proposed change is no longer relevant

PR #14, opened 2018-10-12
 Update kotlinx.coroutines to 1.0.0-RC

qwwdfsad:
...add some review comments in forms of TODO

Do not merge before 1.0.0-RC release.
I've fastly skimmed through API and here are things that caught my eye, but take it with a grain of salt because I haven't evaluated real usages of API.

- package is `kotlinx.coroutines.io`, but folder structure is `kotlinx.coroutines.experimental.io`
- it's good idea to rename package to `kotlinx.io.coroutines`to have a common prefix and avoid clashes with `koltinx.coroutines`. Also, it will help to flatten folders structure
- there is a lot of unused code/API. It should be either documented/tested, or removed (in release different from 0.0.1)
- `IoBuffer` does not have `copy` method as it stated in the documentation. Also, it is not a buffer :) `A read-write facade to actual buffer`, so this part of mental model can be revisited. Should this API be exposed as `public`?
- Naming is inconsisent, e.g. `IoBuffer`, but `ByteChannelSequentialJVM` and `decodeUTF8` (my vote for `decodeUtf8`).
- Some copy-paste between JS and Native
- Some API is prone to overflows. E.g. `fun ByteReadPacket.readBytes(n: Int = remaining.coerceAtMostMaxInt())` will break its contract if packet is larger than 2 GBs.
- Some API (such as `resetForWrite(limit: Int)`) doesn't check for `limit > 0`, but checks that `limit <= writeBuffer.capacity()`. (also, these checks are duplicated by java BB API)
- `@Deprecated(level = ERROR)` in interface seems like API flaw. E.g. what if I want to implement `DevNullInput` in my serialization benchmarks? At least provide a default implementation for these methods
- `Charset` is not documented
- `addSuppressedInt` is probably for `addSuppressedInternal`? 
- `public fun readUTFUntilDelimiterToSlow1`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,8 +1,6 @@
 buildscript {
     repositories {
         if (kotlin_version.contains("eap") || kotlin_version.contains("-rc-") || kotlin_version.contains("-M")) {
-            maven { url 'http://dl.bintray.com/kotlin/kotlin-eap-1.1' }
-            maven { url 'http://dl.bintray.com/kotlin/kotlin-eap-1.2' }
             maven { url 'http://dl.bintray.com/kotlin/kotlin-eap' }
             maven { url 'http://dl.bintray.com/kotlin/kotlin-dev' }
         }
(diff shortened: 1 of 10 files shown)


cy6erGn0m:
Merged manually except coroutines version (as far as there is no 1.0.0-RC released yet). I'll keep it open for a while to see your comments.

PR #20, opened 2018-11-14
Reduce internal dependencies 

cy6erGn0m:


kotlinx-coroutines-io/src/main/kotlin/kotlinx/coroutines/experimental/io/Coroutines.kt
```diff
@@ -128,9 +106,32 @@ fun writer(
     }
 }
 
-internal class ReaderCoroutine(context: CoroutineContext, channel: ByteChannel) :
-    ByteChannelCoroutine(context, channel), ReaderJob, ReaderScope
+/**
+ * @param S not exactly safe (unchecked cast is used) so should be [ReaderScope] or [WriterScope]
+ */
+private fun <S : CoroutineScope> CoroutineScope.launchChannel(context: CoroutineContext,
```
e5l: code style

kotlinx-coroutines-io/kotlinx-coroutines-io-jvm/src/main/kotlin/kotlinx/coroutines/experimental/io/internal/CancellableReusableContinuation.kt
```diff
@@ -103,32 +88,33 @@ internal class MutableDelegateContinuation<T : Any> : Continuation<T>, Dispatche
...
+
             if (job.isActive) {
                 handler = h
             }
         }
 
         override fun invoke(cause: Throwable?) {
-            this@MutableDelegateContinuation.handler.compareAndSet(this, null)
-            dispose()
+            this@CancellableReusableContinuation.jobCancellationHandler.compareAndSet(this, null)
```
qwwdfsad: shouldn't it be protected with if?
```
if (compareAndSet) {
  ... invoke handler and resume
}
```

kotlinx-coroutines-io/src/main/kotlin/kotlinx/coroutines/experimental/io/Coroutines.kt
```diff
@@ -128,9 +106,32 @@ fun writer(
...
+            channel.attachJob(coroutineContext[Job]!!)
+        }
+        try {
+            @Suppress("UNCHECKED_CAST")
+            block(ChannelScope(this, channel) as S)
+        } catch (t: Throwable) {
+            channel.cancel(t)
+            throw CancellationException("Cancelled due to error $t")
+        } finally {
+            channel.close()
```
qwwdfsad: If `block` spawns children, then the channel can be closed before they complete.
Is it intended behaviour?

kotlinx-coroutines-io/src/main/kotlin/kotlinx/coroutines/experimental/io/Coroutines.kt
```diff
@@ -128,9 +106,32 @@ fun writer(
...
+    val job = launch(context) {
+        if (attachJob) {
+            channel.attachJob(coroutineContext[Job]!!)
+        }
+        try {
+            @Suppress("UNCHECKED_CAST")
+            block(ChannelScope(this, channel) as S)
+        } catch (t: Throwable) {
+            channel.cancel(t)
+            throw CancellationException("Cancelled due to error $t")
```
qwwdfsad: Why are you not using t as `cause`?

kotlinx-coroutines-io/src/main/kotlin/kotlinx/coroutines/experimental/io/Coroutines.kt
qwwdfsad: Same about `cancel` call

qwwdfsad:
I am still unhappy about `$foo$` methods.

They are not errors, but they still make core API surface unimplementable and give you a feeling that you have no control over how to implement it or what's happening.

Why a discussed solution with `peek + skip` is not applicable for 1.0? I'd rather not to have these methods in public API at all, because later the only option to remove them is to introduce the whole new interface

cy6erGn0m:
@qwwdfsad Because it is too risky to rewrite everything right now. So basically users shouldn't implement base interfaces but extend base abstract class. This could be fixed in later major releases bumped to next ktor major release.

kotlinx-coroutines-io/kotlinx-coroutines-io-jvm/src/main/kotlin/kotlinx/coroutines/experimental/io/internal/CancellableReusableContinuation.kt
cy6erGn0m: No, it shouldn't. Will rewrite to clarify it

kotlinx-coroutines-io/src/main/kotlin/kotlinx/coroutines/experimental/io/Coroutines.kt
cy6erGn0m: You are right, perhaps need to do it in regular `invokeOnCompletion`, thanks

kotlinx-coroutines-io/src/main/kotlin/kotlinx/coroutines/experimental/io/Coroutines.kt
cy6erGn0m: To avoid double cancellation that caused strange effects in the past because of circular job dependency graph.  Not exactly sure about it

cy6erGn0m: Rewritten so exceptions are not caught at all but handled in `invokeOnCompletion`. Potential double cancellation handling is your problem from this point ;-)

PR #21, opened 2018-11-20
Adds a cpointer constructor to native IoBuffer

SalomonBrys:
so that IoBuffer can be used to read AND write on a memory chunk.

### DIFF
--- a/kotlinx-io-native/src/main/kotlin/kotlinx/io/core/IoBufferNative.kt
+++ b/kotlinx-io-native/src/main/kotlin/kotlinx/io/core/IoBufferNative.kt
@@ -17,6 +17,8 @@ actual class IoBuffer internal constructor(
 ) : Input, Output {
     internal var refCount = 1
 
+    constructor(content: CPointer<ByteVar>, contentCapacity: Int) : this(content, contentCapacity, null)
+
     internal var readPosition = 0
     internal var writePosition = 0
     private var limit = contentCapacity


cy6erGn0m:
I remember your request about it. Unfortunately there are cases when a custom IoBuffer could leak into the default buffer pool that need to be fixed first.

SalomonBrys:
This feature exists in IoBuffer JVM, so I think this should be in IoBuffer Native.
Is this custom buffer leak is handled in JVM ?

I've had to build my own custom version of Kotlinx.io to develop the next component of the Kodein Framework Kodein-DB because it cannot work without it.

cy6erGn0m:
Ah I see, actually the issue has limited impact as one usually is unable to construct packet and channel from custom buffers. 

Could you please explain why do you need cpointer here? does it point to something memory mapped?

SalomonBrys:
tl;dr: I read *and* write on the same IoBuffer.

Kodein-DB is based on LevelDB, a Google C/C++ key-value datastore, on which it adds the concepts of kinds, indices, etc.
To that end, keys are composed and contains various informations.

Now, Kodein-DB tries very hard to limit as much as possible data duplication, so for example when modifying a key, the key is loaded inside an IoBuffer (via this new constructor), than the key is modified *in place*.
The same thing happens when creating indices key-values.

Finally, for "simple" writes, I need to ensure that everything happens on the same contiguous memory chunk as LevelDB do not support having keys or values on non-contiguous memory.
Most of the time, Kodein-DB can predict the length of the value and allocate a buffer of the right size.

PR #36, opened 2019-02-05
Fix readRemaining

e5l:


cy6erGn0m:
Thanks!

cy6erGn0m:
The fix is available in dev build 0.1.5-dev-4

jrocharodrigues:
I had this issue ( Kotlin/kotlinx-io#360 ) when using serialization and ktor. Forced the use of kotlinx-io-0.1.5-dev-4 and org.jetbrains.kotlinx:kotlinx-coroutines-io-0.1.5-dev-4 and confirm this issue is solved. 

However when using Ktor MockEngine to mock the the Json answer i got an error similar to the previous one i reported (the limit is different though) `Invalid JSON at 4088: Unexpected end in string`

Could this error be related to this issue, or should i open a new on for Ktor?

Thanks in advance

e5l:
It looks like the separate issue from kx.io could you file an issue with reproducer or data sample?

jrocharodrigues:
Ok, i have opened an issue in ktor, but i will close it, and open one in kx.io

jrocharodrigues:
Done: https://github.com/Kotlin/kotlinx-io/issues/38

PR #51, opened 2019-05-31
fix typo `if` instead of `of`

robstoll:


qwwdfsad:
We are completely reworking current implementation

PR #65, opened 2019-10-31
Experimental propagation

qwwdfsad:


core/commonMain/src/kotlinx/io/Input.kt
```diff
@@ -114,100 +114,61 @@ public abstract class Input : Closeable {
...
-     * Reads an [UInt] from this Input.
-     *
-     * @throws EOFException if no more bytes can be read.
-     */
-    public fun readUInt(): UInt = readPrimitive(
-        4, { buffer, offset -> buffer.loadUIntAt(offset) }
-    ) { it.toUInt() }
+        4,
+        { buffer, offset -> buffer.loadIntAt(offset) },
+        { it.toInt() })
```
e5l: Codestyle

core/commonMain/src/kotlinx/io/Input.kt
```diff
@@ -319,8 +266,6 @@ public abstract class Input : Closeable {
         previewBytes?.close()
     }
 
-    public companion object {}
```
e5l: let's live the companion as it is for now

core/commonMain/src/kotlinx/io/Input.kt
qwwdfsad: I've left each lambda on its own line deliberately to improve readability.

PR #73, opened 2019-11-14
Read lines

qwwdfsad:


core/commonMain/src/kotlinx/io/Outputs.kt
```diff
@@ -0,0 +1,49 @@
...
+    /** @suppress */
+    override fun close() {
+        // Nothing by default
+    }
+
+    @UseExperimental(ExperimentalStdlibApi::class)
+    private fun ensureCapacity(length: Int) {
+        if (array.size < size + length) {
+            val minCapacity = size + length
+            val powOf2 =
```
e5l: Consider extracting function here

core/commonMain/src/kotlinx/io/text/TextInputObsolete.kt
```diff
@@ -64,34 +65,34 @@ fun Input.readUTF8LineTo(out: Appendable) {
...
-private inline fun Input.decodeUTF8Chars(consumer: (Char) -> Boolean): Int {
+private inline fun Input.decodeUtf8Chars(consumer: (Char) -> Boolean): Int {
     var byteCount = 0
     var value = 0
-    var state = STATE_UTF8
+    var state = STATE_UTF_8
     var count = 0
 
-    while (state != STATE_FINISH) {
+    while (state != STATE_FINISH && !eof()) {
```
e5l: Why we expect `eof` here?

core/commonMain/src/kotlinx/io/text/TextInputObsolete.kt
```diff
@@ -37,25 +38,25 @@ fun Input.readUTF8StringTo(out: Appendable, length: Int): Int {
 /**
  * @throws MalformedInputException
  */
-fun Input.readUTF8Line(): String = buildString {
-    readUTF8LineTo(this)
-}
+//fun Input.readUtf8Line(): String = buildString {
```
e5l: Consider removing this method

core/commonMain/src/kotlinx/io/text/TextInput.kt
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/**
+ * Iterates through each line of the input, calls [action] for each line read
+ * and closes the input when it is exhausted.
+ *
+ * @throws EOFException if the input is already exhausted
+ */
+public inline fun Input.forEachUtf8Line(action: (String) -> Unit) {
+    if (eof()) throw EOFException("Unexpected EOF while reading UTF-8 line")
```
e5l: Consider using `Input.checkExhausted()` here.

e5l: Throwing on empty input looks strange to me. Should we consider the use cases first?

core/commonMain/src/kotlinx/io/Outputs.kt
qwwdfsad: `ensureCapacity` is already a separate method

core/commonMain/src/kotlinx/io/text/TextInputObsolete.kt
qwwdfsad: To allow reading strings that are not terminated with a line break.
Frankly speaking, we should also check that eof didn't happen in the middle of the code point, but I decided not to touch the implementation of text decoding at all for now

core/commonMain/src/kotlinx/io/text/TextInput.kt
qwwdfsad: Good point, `forEachUtf8Line` should be a no-op for empty and exhausted inputs.
My initial idea seems to be flawed

core/commonMain/src/kotlinx/io/Outputs.kt
```diff
@@ -0,0 +1,49 @@
...
+                if (minCapacity >= 1 shl 30) minCapacity else (1 shl (32 - (minCapacity - 1).countLeadingZeroBits()))
```
EdwarDDay: Can
```
(1 shl (32 - (minCapacity - 1).countLeadingZeroBits()))
```
be replaced by
```
(minCapacity - 1).takeHighestOneBit() shl 1
```
?
This looks easier to understand IMHO.

PR #74, opened 2019-11-14
Introduce closeSource method to Output

qwwdfsad:
* Make Output.close final
    * Flush buffered data before closing the underlying source in Output.close because it is the only reasonable behaviour
    * Properly handle exceptions during termination sequence

### DIFF
--- a/core/commonMain/src/kotlinx/io/Output.kt
+++ b/core/commonMain/src/kotlinx/io/Output.kt
@@ -88,6 +88,31 @@ abstract class Output(bufferSize: Int = DEFAULT_BUFFER_SIZE) : Closeable {
     }
 
     /**
+     * Closes the underlying source of data used by this output.
+     * This method is invoked once the output is [closed][close].
+     */
+    protected abstract fun closeSource()
+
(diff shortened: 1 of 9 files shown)


core/commonMain/src/kotlinx/io/Output.kt
```diff
@@ -87,6 +87,33 @@ abstract class Output(bufferSize: Int = DEFAULT_BUFFER_SIZE) : Closeable {
...
+    /**
+     * Closes the current output, flushing all buffered data to the underlying source
+     * and [closing][closeSource] it
+     */
+    public final override fun close() {
+        val flushException = runCatching { flush() }.exceptionOrNull()
+        val closeException = runCatching { closeSource() }.exceptionOrNull()
+        if (flushException !== null) {
+            if (closeException !== null) {
+                if (closeException !== flushException) {
```
e5l: `if (104 && 105)`

e5l: btw. consider extracting this to a helper method

PR #78, opened 2019-11-25
Introduce Input.limit

qwwdfsad:


core/commonMain/src/kotlinx/io/Inputs.kt
```diff
@@ -10,3 +12,16 @@ public fun ByteArrayInput(source: ByteArray, startIndex: Int = 0, endIndex: Int
     }
     return kotlinx.io.internal.ByteArrayInput(source, startIndex, endIndex)
 }
+
+/**
+ * Wraps an input, limiting the number bytes which can be read up to the given [limit].
```
e5l: It's unclear what will happen in the end.

PR #95, opened 2020-03-05
Introduce interoperability with java.io

qwwdfsad:


core/jvmMain/src/kotlinx/io/internal/JavaInterop.kt
```diff
@@ -0,0 +1,92 @@
+package kotlinx.io.internal
+
+import kotlinx.io.*
+import kotlinx.io.buffer.*
+import java.io.*
+
+internal class InputStreamFromInput(private val input: Input) : InputStream() {
+    override fun read(): Int {
+        if (input.exhausted()) {
+            return -1
```
e5l: consider lifting return(in the PR)

core/jvmMain/src/kotlinx/io/internal/JavaInterop.kt
```diff
@@ -0,0 +1,92 @@
...
+internal class InputFromInputStream(private val inputStream: InputStream) : Input() {
+    override fun closeSource() {
+        inputStream.close()
+    }
+
+    override fun fill(buffer: Buffer, startIndex: Int, endIndex: Int): Int {
+        // Zero-copy attempt
+        if (buffer.buffer.hasArray()) {
+            return inputStream.read(buffer.buffer.array(), startIndex, endIndex - startIndex)
+                .coerceAtLeast(0) // -1 when IS is closed
```
e5l: please consider avoiding index calculations in arguments(especially with methods or corner cases)

core/jvmMain/src/kotlinx/io/internal/JavaInterop.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        for (i in startIndex until endIndex) {
+            val byte = inputStream.read()
```
e5l: Could we preallocate a ByteArray here? Maybe we can make it as lazy field.

PR #100, opened 2020-03-12
Bytes rework

e5l:
- make bytes internal
- introduce BytesInput & BytesOutput


benchmarks/commonMain/src/kotlinx/io/BytesInputReadBenchmark.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.benchmark.*
+import kotlinx.io.bytes.*
+
+/**
+ * Benchmark with prebuilt ByteInput
+ */
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(BenchmarkTimeUnit.MICROSECONDS)
+class BytesInputReadBenchmark {
```
qwwdfsad, line 12: If this benchmark is measuring preview overhead similarly to Okio, could you please add Okio to this benchmark?

benchmarks/commonMain/src/kotlinx/io/BytesInputReadBenchmark.kt
```diff
@@ -0,0 +1,67 @@
...
+    lateinit var input: Input
+
+    @Param("2097152")
+    open var size: Int = 42
```
qwwdfsad, line 16: open is redundant

core/commonMain/src/kotlinx/io/Output.kt
```diff
@@ -14,7 +14,10 @@ import kotlinx.io.pool.*
 public abstract class Output(
     protected val pool: ObjectPool<Buffer> = DefaultBufferPool.Instance
 ) : Closeable {
-    internal val bufferPool: ObjectPool<Buffer> get() = pool
+    /**
+     * Buffer pool used by this [Input] for copy and cache operations.
+     */
+    val bufferPool: ObjectPool<Buffer> get() = pool
```
qwwdfsad, line 20: 1) Why do we need it `public`?
2) explicit `public`
3) What's the difference between `pool` and `bufferPool`? If we have to make it public, why don't we stick with a single getter?

core/commonMain/src/kotlinx/io/bytes/BytesOperations.kt
```diff
@@ -0,0 +1,25 @@
+package kotlinx.io.bytes
+
+import kotlinx.io.*
+import kotlinx.io.buffer.*
+import kotlin.contracts.*
+
+/**
+ * Create [BytesInput] with content from [block] using specified [bufferSize].
+ */
+fun buildInput(bufferSize: Int = DEFAULT_BUFFER_SIZE, block: BytesOutput.() -> Unit): BytesInput {
```
qwwdfsad, line 10: explicit public

core/commonMain/src/kotlinx/io/bytes/Bytes.kt
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Index of the first [Buffer] in [buffers].
+     */
+    var head: Int = 0
+        private set
+
+    /**
+     * Index of the last [Buffer] in [buffers].
+     */
+    var tail: Int = 0
```
qwwdfsad, line 26: should also have private set

core/commonMain/src/kotlinx/io/bytes/BytesInput.kt
```diff
@@ -0,0 +1,30 @@
+package kotlinx.io.bytes
+
+import kotlinx.io.*
+import kotlinx.io.buffer.*
+
+/**
+ * In memory byte source created from [BytesOutput] or by using [buildInput].
+ *
+ * [BytesInput] isn't pooled so it's safe to throw it away.
```
qwwdfsad, line 9: `throw it away` -> `not to close it`.
It's better to be explicit in the documentation

core/commonMain/src/kotlinx/io/bytes/BytesInput.kt
```diff
@@ -0,0 +1,30 @@
...
+) : Input(bytes, unmanagedPoolOfBuffers(bufferSize)) {
+
+    /**
+     * Count of remaining bytes in [BytesInput].
+     */
+    public val remaining: Int
+        get() = remainingCacheSize()
+
+    override fun fill(buffer: Buffer, startIndex: Int, endIndex: Int): Int {
+        return 0
```
qwwdfsad, line 23: This one should never be called, so it's better to throw here in order to fail fast

core/commonMain/src/kotlinx/io/bytes/BytesInput.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import kotlinx.io.*
+import kotlinx.io.buffer.*
+
+/**
+ * In memory byte source created from [BytesOutput] or by using [buildInput].
+ *
+ * [BytesInput] isn't pooled so it's safe to throw it away.
+ */
+public class BytesInput internal constructor(
```
qwwdfsad, line 11: The problem with `BytesInput` and `BytesOutput`: it's impossible to understand why is it needed, what problem it is intended to solve and how to use it.
I would recommend you to take `kotlinx.coroutines` approach to the documentation (e.g. take a look at `channelFlow` function or `StateFlow`)

benchmarks/commonMain/src/kotlinx/io/BytesInputReadBenchmark.kt
qwwdfsad: Also, please properly document what this benchmark is expected to measure and why it is important.
Otherwise, we would end up in a situation pretty similar to what we have now :)

core/commonMain/src/kotlinx/io/bytes/BytesOperations.kt
qwwdfsad: And the same comment about usefulness of this primitive that is not reflected in the documentation

core/commonMain/src/kotlinx/io/bytes/BytesOperations.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    return BytesOutput(bufferSize).apply {
+        block()
+    }.createInput()
+}
+
+/**
+ * Read [BytesInput] of fixed [size].
+ */
+public fun Input.readBytesInput(size: Int): BytesInput = buildInput {
```
qwwdfsad, line 23: Such family of operators is typically named with prefix `to` in Kotlin.
Please stick with an official naming scheme

core/commonMain/src/kotlinx/io/bytes/BytesOutput.kt
```diff
@@ -0,0 +1,46 @@
+package kotlinx.io.bytes
+
+import kotlinx.io.*
+import kotlinx.io.buffer.*
+
+/**
+ * Create unlimited [Output] stored in memory.
```
qwwdfsad, line 7: Here should be documentation to the class, not to the constructor immediately

core/commonMain/src/kotlinx/io/bytes/BytesOutput.kt
```diff
@@ -0,0 +1,46 @@
...
+ * In advance to [Output] you can check [size] and create [BytesInput] with [createInput].
+ * [BytesOutput] isn't using any pools and shouldn't be closed.
+ */
+public class BytesOutput(
+    private val bufferSize: Int = DEFAULT_BUFFER_SIZE
```
qwwdfsad, line 12: could you please elaborate why `BytesInput` constructor is internal, while this one is not?

benchmarks/commonMain/src/kotlinx/io/BytesInputReadBenchmark.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    @Setup
+    fun setup() {
```
qwwdfsad, line 19: Have you managed to find real-world benchmarks previous IO was optimized against? Maybe ask @cy6erGn0m while we are here?

core/commonMain/src/kotlinx/io/Output.kt
e5l: 1. We need it `public` to use in transformers.
3. I'll keep the one from the constructor.

benchmarks/commonMain/src/kotlinx/io/BytesInputReadBenchmark.kt
e5l: They are in a separate branch.

core/commonMain/src/kotlinx/io/bytes/BytesOutput.kt
e5l: You can't create `BytesInput` from internal `Bytes` in public, and can create empty `BytesOutput` for write.

core/commonMain/src/kotlinx/io/bytes/BytesInput.kt
e5l: Thanks. Will do.

core/commonMain/src/kotlinx/io/bytes/BytesInput.kt
e5l: It will be called at the end of the input. We can't indicate `eof` without last fill.

e5l:
closed as obsolete

PR #101, opened 2020-03-12
Add flush flag & leak detector

e5l:
- Add flush return flag to indicate if we can reuse the buffer
- introduce leak detector


core/commonMain/src/kotlinx/io/internal/LeakDetectingPool.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import kotlinx.io.buffer.*
+import kotlinx.io.pool.*
+
+/**
+ * The pool that tracks every borrowed object and check if it belongs to this pool.
+ *
+ * It also verifies if this pool is empty at close.
+ */
+internal class LeakDetectingPool(private val size: Int = DEFAULT_BUFFER_SIZE) : ObjectPool<Buffer> {
```
qwwdfsad: Looks like it should belong to test module

core/commonTest/src/kotlinx/io/InputPrimitivesTest.kt
```diff
@@ -14,13 +14,15 @@ class InputPrimitivesTest {
         limit: Int, seed: Long = 0L, body: Input.() -> Unit
     ) = prefetchSizes.forEach { prefetchSize ->
         bufferSizes.forEach { size ->
+            val input = sequentialLimitedInput(fetchSizeLimit, pool, limit, seed)
```
qwwdfsad: `use`?

core/commonMain/src/kotlinx/io/internal/LeakDetectingPool.kt
e5l: Yep, but we can't share them across test modules

core/commonMain/src/kotlinx/io/internal/LeakDetectingPool.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    override fun close() {
+        val liveBuffers = allocated.size - removed.size
+        check(liveBuffers == 0) {
+            "Memory leak: $liveBuffers buffers still in use. Total allocated: ${allocated.size}"
+        }
+    }
+
+    companion object {
+        private var poolCounter: Byte = 0
```
qwwdfsad: unused

e5l:
closed as obsolete

Issue #107, opened 2020-03-31
readByteArray is broken


altavir:
https://github.com/Kotlin/kotlinx-io/blob/2b54f00af7346ed9ff52213b33821e2d588ea3e4/core/commonMain/src/kotlinx/io/InputOperations.kt#L64-L65

This two statements together mean that returned array is always empty and pointer not advanced. I recommend adding tests for that.

altavir:
Even after my fix, the method is still broken. By contract, this method:
https://github.com/Kotlin/kotlinx-io/blob/fe899e0e2cbef86cbf7354f7fd774807b341cf34/core/commonMain/src/kotlinx/io/InputOperations.kt#L66-L71
should return the position after read (I don't know how it meant to know the position - it is private), but returns the number of bytes being read, so it works only when we are reading at the beginning of new buffer. Any suggestions?

fzhinkin:
We're rebooting the kotlinx-io development (see https://github.com/Kotlin/kotlinx-io/issues/131), all issues related to the previous versions will be closed. Consider reopening it if the issue remains (or the feature is still missing) in a new version.

PR #108, opened 2020-03-31
[Not for merge] Binaries

altavir:
* Introduce Binary (multi-read independent byte block) and implementation for Byte array and JVM Path.
* Tests for them
* Fix for #107
* Add browser target

Some node tests are failing.

### DIFF
--- a/core/commonMain/src/kotlinx/io/InputOperations.kt
+++ b/core/commonMain/src/kotlinx/io/InputOperations.kt
@@ -1,7 +1,7 @@
 package kotlinx.io
 
 import kotlinx.io.buffer.*
-import kotlin.math.*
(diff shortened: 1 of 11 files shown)


core/commonMain/src/kotlinx/io/InputOperations.kt
```diff
@@ -61,7 +61,7 @@ public fun Input.readByteArray(size: Int): ByteArray {
     checkSize(size)
 
     val result = ByteArray(size)
-    var position = size
+    var position = 0
```
altavir, line 64: This line fixes #107

altavir:
Oops, wrong file commited as fix for #107 . Added comment for the correct line.

PR #111, opened 2020-04-29
Fast EOF

qwwdfsad:


e5l:
Hi @qwwdfsad, thanks for the PR. As mentioned before, this change won't prevent filling the internal `Input` buffer. Could you add some measurements to compare the change?

qwwdfsad:
Yeah, that's why I haven't touched our primitives like `copyTo`.
This change is more about readability and the code being straightforward, not about performance

e5l:
LGTM

core/commonMain/src/kotlinx/io/Input.kt
```diff
@@ -214,7 +214,9 @@ public abstract class Input : Closeable {
      * performance-critical code.
      */
     @Suppress("NOTHING_TO_INLINE")
```
e5l, line 216: Suppress can be removed.

PR #120, opened 2021-02-06
Fix writeFully argument validation in Builder.kt

BenWoodworth:
```kt
override fun writeFully(src: IntArray, offset: Int, length: Int) {
    require(length >= 0) { "length shouldn't be negative: $length" }
    require(offset + length < src.lastIndex) { "offset ($offset) + length ($length) >= src.lastIndex (${src.lastIndex})" }
    // ...
```

The second `require` statement was incorrect in the writeFully methods. (e.g. for `src=[1], offset=0, length=1`, you'll get `require(0 + 1 < 0)`)

I also added validation to make sure offset is non-negative, like in other places in the repo.


BenWoodworth:
I'm not sure why it's failing CI checks (compiles fine and passes tests on my machine), but let me know if any changes need to be made

qwwdfsad:
Closing as "won't fix", please see https://github.com/Kotlin/kotlinx-io#obsolete-warning

PR #129, opened 2023-06-01
Add benchmarks suite

fzhinkin:
Added kotlinx-io benchmarks suite.

The suite aimed to track how the performance of the library changes between releases and also to validate the performance on platforms other than JVM.

Currently, only benchmarks covering the very core part of an API were added as that API's part likely won't change in the near future.

The suite doesn't facilitate comparing the performance with other libraries (Okio, ktor-io), but for Okio it could be done by simply replacing import statements in benchmarks from`kotlinx.io` to `okio`, thanks to API's compatibility.

### DIFF
--- a/benchmarks/src/commonMain/kotlin/BufferOps.kt
+++ b/benchmarks/src/commonMain/kotlin/BufferOps.kt
@@ -290,14 +290,16 @@ open class Utf8StringBenchmark: BufferRWBenchmarkBase() {
 
     @Benchmark
     fun kio(): String {
+        val s = buffer.size
         buffer.writeUtf8(string)
-        return buffer.readUtf8(length.toLong())
+        return buffer.readUtf8(buffer.size - s)
     }
 
     @Benchmark
     fun okio(): String {
+        val s = okioBuffer.size
         okioBuffer.writeUtf8(string)
-        return okioBuffer.readUtf8(length.toLong())
+        return okioBuffer.readUtf8(okioBuffer.size - s)
     }
 }
 
--- a/benchmarks/src/commonMain/kotlin/BufferOps.kt
+++ b/benchmarks/src/commonMain/kotlin/BufferOps.kt
@@ -12,7 +12,7 @@ import kotlinx.io.*
 abstract class BufferRWBenchmarkBase {
     // Buffers are implemented as list of segments, as soon as a segment is empty
     // it will be unlinked. By reading all previously written data, a segment will be
-    // cleared and recycled and the next time we will try to write data a new segment
+    // cleared and recycled, and the next time we will try to write data a new segment
     // will be requested from the pool. Thus, without having some data in-flight we will
     // benchmark not only read/write ops performance, but also segments allocation/reclamation.
     @Param("128")
@@ -28,7 +28,7 @@ abstract class BufferRWBenchmarkBase {
         buffer.write(padding)
     }
 
-    @TearDown()
+    @TearDown
     fun clearBuffers() {
         buffer.clear()
     }
@@ -136,7 +136,7 @@ open class Utf8CodePointBenchmark: BufferRWBenchmarkBase() {
     @Param("1", "2", "3")
     var bytes: Int = 0
 
-    var codePoint: Int = 0
+    private var codePoint: Int = 0
 
     @Setup
     fun setupCodePoints() {
@@ -155,7 +155,8 @@ open class Utf8CodePointBenchmark: BufferRWBenchmarkBase() {
     }
 }
 
-// This benchmark is based on https://raw.githubusercontent.com/square/okio/master/okio/jvm/jmh/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
+// This benchmark is based on Okio benchmark:
+// https://raw.githubusercontent.com/square/okio/master/okio/jvm/jmh/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
 open class Utf8StringBenchmark: BufferRWBenchmarkBase() {
     private val strings = mapOf(
         "ascii" to ("Um, I'll tell you the problem with the scientific power that you're using here, "
--- a/benchmarks/src/commonMain/kotlin/BufferOps.kt
+++ b/benchmarks/src/commonMain/kotlin/BufferOps.kt
@@ -10,11 +10,13 @@ import kotlinx.io.*
 
 @State(Scope.Benchmark)
 abstract class BufferRWBenchmarkBase {
-    // Buffers are implemented as list of segments, as soon as a segment is empty
+    // Buffers are implemented as a list of segments, as soon as a segment is empty
     // it will be unlinked. By reading all previously written data, a segment will be
-    // cleared and recycled, and the next time we will try to write data a new segment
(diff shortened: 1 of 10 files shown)


benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -0,0 +1,331 @@
...
+
+open class Utf8LineStrictBenchmark: Utf8LineBenchmarkBase() {
+    @Benchmark
+    fun benchmark(): String {
+        buffer.writeUtf8(string)
+        return buffer.readUtf8LineStrict()
+    }
+}
+
+private const val INDEX_OF_TARGET_VAL: Byte = 1
```
shanshin, line 284: Maybe call something like "Indexed Value"?
Actually, the byte value itself is stored here, and not the index by which it is located.

benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -0,0 +1,331 @@
...
+        var paddingSize = 0L
+        if (spanOverMultipleSegment) {
+            paddingSize = 8192L - dataSize / 2L
+        }
+        val padding = ByteArray(paddingSize.toInt())
+        buffer.write(padding).write(array).skip(paddingSize)
+    }
+
+    @Benchmark
+    fun benchmark(): Long = buffer.indexOf(INDEX_OF_TARGET_VAL)
```
shanshin, line 314: Do I understand correctly that this test involves searching for a value only within one segment?

benchmarks/src/commonMain/kotlin/PeekBenchmark.kt
```diff
@@ -0,0 +1,48 @@
...
+package kotlinx.io.benchmarks
+
+import kotlinx.io.*
+import kotlinx.benchmark.*
+
+@State(Scope.Benchmark)
+abstract class PeekBenchmark {
+    protected val buffer = Buffer()
+
+    @Param("0", "8191")
```
shanshin, line 15: 8191 - is a magic value?

benchmarks/src/commonMain/kotlin/SegmentPoolBenchmark.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import kotlinx.io.*
+import kotlinx.benchmark.*
+
+@State(Scope.Benchmark)
+open class SegmentPoolBenchmark {
+    private val buffer = Buffer()
+
+    @Benchmark
+    fun acquireReleaseCycle() {
```
shanshin, line 16: If this test is run in parallel with the others, will they affect each other (for example, if common static fields are used)?

benchmarks/src/jvmMain/kotlin/SegmentPoolBenchmarkMT.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import kotlinx.io.*
+import kotlinx.benchmark.*
+import org.openjdk.jmh.annotations.Group
+import org.openjdk.jmh.annotations.GroupThreads
+
+@State(Scope.Benchmark)
+open class SegmentPoolBenchmarkMT {
+    private fun testCycle(): Buffer {
+        val buffer = Buffer().writeByte(0)
```
shanshin, line 16: Maybe we should add cycles here to increase the probability of concurrency access to segments?

However, in this case, the test time may become unstable.

benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -0,0 +1,331 @@
...
+import kotlinx.io.*
+
+@State(Scope.Benchmark)
+abstract class BufferRWBenchmarkBase {
+    // Buffers are implemented as list of segments, as soon as a segment is empty
+    // it will be unlinked. By reading all previously written data, a segment will be
+    // cleared and recycled, and the next time we will try to write data a new segment
+    // will be requested from the pool. Thus, without having some data in-flight we will
+    // benchmark not only read/write ops performance, but also segments allocation/reclamation.
+    @Param("128")
```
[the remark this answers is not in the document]
fzhinkin: Mostly random. I checked other values and there were almost no difference in resulting performance. I left it as a parameter to easily check how it will affect performance in the future.

benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -0,0 +1,331 @@
...
+    var minGap: Int = 0
+
+    protected val buffer = Buffer()
+
+    protected open fun padding(): ByteArray = ByteArray(minGap)
+
+    @Setup
+    fun setupBuffers() {
+        val padding = padding()
```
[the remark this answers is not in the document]
fzhinkin: The padding is taken into account because the first read from buffer will consume its part. That's why the padding depends on particular test - it has to be meaningful.

benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -0,0 +1,331 @@
...
+
+    @Benchmark
+    fun benchmark(): Long = buffer.indexOf(INDEX_OF_TARGET_VAL)
+}
+
+@State(Scope.Benchmark)
+open class BufferGetBenchmark {
+    private val buffer = Buffer()
+
+    @Param("0", "8193")
```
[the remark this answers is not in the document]
fzhinkin: Exactly. It corresponds to the second byte in the second segment.

benchmarks/src/commonMain/kotlin/BufferOps.kt
fzhinkin: The intent was to name it like "`indexOf`'s target value", but now I see it didn't work out. :)
Sure, I'll peek a different name.

benchmarks/src/commonMain/kotlin/PeekBenchmark.kt
fzhinkin: It's an index of the last byte within the first segment. The idea was to hit segments' border in multi-byte benchmarks.
I'll add a comment regarding its meaning.

qwwdfsad (approved):
Nice one!

Benchmarks are more or less self-descriptive and it's clear what they are targeing.
Would be nice to add a bit more comments around magic numbers (or even use constant if it's a segment size etc.) and non-trivialities (padding, indents in decimals benchmarks etc.).

benchmarks/src/commonMain/kotlin/SegmentPoolBenchmark.kt
fzhinkin: This one will not be executed concurrently unless the user will explicitly specify the opposite. And in concurrent case it may fail due to lack of synchronization.

Not sure if there is a way to confine benchmark's state to a single thread using `kotlinx.benchmark`'s annotations (like it could be done using JMH's `@Scope(Thread)`.

benchmarks/src/jvmMain/kotlin/SegmentPoolBenchmarkMT.kt
fzhinkin: Under the hood, a runner invokes benchmarked methods in a loop, so I don't think it'll have an impact on overall benchmark execution.

shanshin (approved):

Issue #132, opened 2023-06-12
kotlinx-io API as a trimmed-down Okio API


fzhinkin:
As a starting point for https://github.com/Kotlin/kotlinx-io/issues/131, `kotlinx-io` API should be implemented as a trimmed-down version for Okio API. The API should include only core IO primitives and allow to build any other functionality on top of it.

Parts of the Okio API that are likely to be removed include:
- select API;
- hash functions API;
- cursor API;
- ByteStrings (will be added back later).

Some of these APIs provide specific functionality required by particular libraries and apps and could be implemented on top of the existing API; others are unsafe and should be redesigned to reduce possible user issues. It's also clear that these APIs were implemented as part of the core Okio library to squeeze additional performance out of it. To make sure that these APIs can be implemented using `kotlinx-io` without the loss of performance, the new APIs will be introduced in the future (or, if there will be no other way to implement removed functionality efficiently, it may be added back to the `kotlinx-io`).

The goal is to provide concise, well-documented and tested API. Core IO primitives API, like sources and sinks, should include as little functionality as possible; everything that could be implemented on top of it should be moved to extension functions or removed.

qwwdfsad:
I would suggest starting with use cases that you are solving with an API (either Okio and/or obsolete Input/Output) to help us shape the overall direction. The same question stands for the implementation -- what exactly you do want to implement?

qwwdfsad:
>I have two primary use-cases

These two use cases will be solvable with the proposed API shape and similar performance characteristics, I do not see issues with the proposed Okio-based design right off the bat.

qwwdfsad:
We are happy to have any feedback about the API, and are keen on addressing known and yet unknown problems, whether they are about API shape, naming or performance characteristics.
We'll do an initial release soon, and I suggest you trying it.

Yet please consider being a bit more constructive -- the feedback "This API is more powerful and it would be nice to address it" is non-actionable for us and we cannot reasonably act on it

Issue #135, opened 2023-06-12
Provide an API for performant bulk or sequential read/writes


fzhinkin:
As stated in https://github.com/Kotlin/kotlinx-io/issues/132, `kotlinx-io` don't include Okio APIs aimed for performant sequential and bulk ops (like, UnsafeCursor).

Such API is required to implement previously removed functionality (like select) efficiently. It's also required for integration with platform-specific APIs (like hash functions implementation using Java's `MessageDigest` API).

This issue claims the intent to provide such functionality, but the actual API will be discussed later.

whyoleg:
Just wanted to mention, that from my perspective this should be implemented taking into account future feature from #131: `different memory types (in addition to byte arrays) Buffers are built upon (like NIO ByteBuffers on JVM);`
Use-case: https://github.com/ktorio/ktor-io/issues/7
Ideally it should be possible to use JDK `MessageDigest`/`Cipher` API for `Buffer` with multiple underlying segments of different types like `ByteBuffer`, `ByteArray`, `MemorySegment` and even Netty `ByteBuf` (which allows to get `ByteBuffer` from it).

fzhinkin:
An implementation will definitely take different memory types into account, otherwise, it'll be hard to make performant reads/writes.

rnett:
I am trying to implement a custom `HashingSink` (using Guava's hash functions, but it doesn't really matter) and found that implementing a custom sink is rather difficult because it's not easy to read (groups of) bytes from a `Buffer` yourself, especially without "consuming" them (so I can forward to the "real" sink).  This makes it very hard to implement custom sinks or sources.  It seems like this issue would be the feature responsible for solving this, is that correct?  If not I'll create a new one.

fzhinkin:
@rnett, that's correct. Once solved, this issue should allow more easily implementing things like `HashingSink`. I'm currently finalizing the design and will post it soon.

fzhinkin:
Here's an update with the details on an API we will implement.

## Intro

In general, there are two groups of cases when existing public API works poorly both performance- and convenience-wise:
- integration with other frameworks/libraries/platform APIs;
- custom extensions.

The first group usually requires taking as much data as possible and passing it down to other API, or, reserving some 
empty space in advance and then supply it to the underlying API to fill it up.
For example, POSIX declares vectorized-io functions, `readv` and `writev`, allowing to read/write multiple chunks of
[...]

PR #136, opened 2023-06-12
A trimmed-down API

fzhinkin:
Reduced `kotlinx-io` API shape by removing all niche and specific APIs.

Moved as many functions as possible to extensions, improved test coverage, and added documentation.

Docs generated for current branch are here: https://fzhinkin.github.io/kotlinx-io-dokka-docs-preview/

What changed compared to the previous preview version, in detail:
- removed select API;
- removed ByteStrings;
- removed hash functions API;
- added unsigned integer types support;
- `Sink::writeByte` and `Sink::writeShort` now accept `Byte` and `Short` instead of `Int`;
- `Sink`, `Source`, and `Buffer` no longer implement ByteChannels on JVM; functions returning wrappers around instances of these classes and implementing ByteChannels were provided instead;
- introduced `DelicateIoApi` annotation, marked `Sink::buffer`, `Sink::emitCompleteSegments`, and `Source::buffer` with it;
- enabled code coverage collection using `Kover`;
- enabled Dokka for documentation;
- turned on explicit API mode.

Some issues were not addressed yet:
- [x] choose and document consistent behavior when it comes to throwing exceptions (https://github.com/Kotlin/kotlinx-io/issues/138);
- [x] choose new naming for read/write methods (https://github.com/Kotlin/kotlinx-io/issues/137).
- [x] readFully behaves inconsistently with other read methods in case of thrown exception (https://github.com/Kotlin/kotlinx-io/issues/139).

What's out of this change's scope and will be resolved later:
- performance;
- ByteStrings;
- replacement for cursor API.

Closes https://github.com/Kotlin/kotlinx-io/issues/132


core/common/src/internal/-RealBufferedSource.kt
```diff
@@ -39,6 +41,7 @@ internal inline fun RealSource.commonRead(sink: Buffer, byteCount: Long): Long {
   return buffer.read(sink, toRead)
 }
 
+@OptIn(DelicateIoApi::class)
 internal inline fun RealSource.commonExhausted(): Boolean {
   check(!closed) { "closed" }
```
qwwdfsad, line 46: Exception message could be a bit more verbose probably

core/jvm/src/Buffer.kt
```diff
@@ -21,557 +21,281 @@
...
 
-  @Throws(EOFException::class)
-  override fun readUtf8LineStrict(limit: Long): String = commonReadUtf8LineStrict(limit)
+  actual override fun close(): Unit = Unit
 
-  @Throws(EOFException::class)
-  override fun readUtf8CodePoint(): Int = commonReadUtf8CodePoint()
+  /**
+   * Returns a human-readable string that describes the contents of this buffer. Typically, this
+   * is a string like `[text=Hello]` or `[hex=0000ffff]`.
```
qwwdfsad, line 116: It would be nice to know if it can be used as a unique identifier (e.g. to compare two arbitrary buffers in an ad-hoc manner)

core/jvm/src/Buffer.kt
```diff
@@ -21,557 +21,281 @@
...
+  public actual fun copy(): Buffer = commonCopy()
 
-  @Throws(EOFException::class)
-  override fun readByteArray(byteCount: Long): ByteArray = commonReadByteArray(byteCount)
+  /**
+   * Returns a deep copy of this buffer.
+   *
+   * This method is equivalent to [copy].
+   */
+  public override fun clone(): Buffer = copy()
```
qwwdfsad, line 127: Do we really want to pursue `Cloneable` here?

core/jvm/src/Buffer.kt
```diff
@@ -21,557 +21,281 @@
...
 import java.nio.ByteBuffer
 import java.nio.channels.ByteChannel
-import java.nio.charset.Charset
-import java.security.InvalidKeyException
-import java.security.MessageDigest
-import javax.crypto.Mac
-import javax.crypto.spec.SecretKeySpec
 
-actual class Buffer : Source, Sink, Cloneable, ByteChannel {
+public actual class Buffer : Source, Sink, Cloneable {
```
qwwdfsad, line 31: Since `Buffer` no longer implements `ByteChannel`, is it possible to have a single `common` implementation now?

core/jvm/test/JvmPlatformTest.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    @Test fun inputStreamSource() {
+        val bais = ByteArrayInputStream(byteArrayOf(0x61))
+        val source = bais.source()
+        val buffer = Buffer()
+        source.read(buffer, 1)
+        assertEquals(buffer.readUtf8(), "a")
+    }
+
+    @Test fun fileSink() {
```
qwwdfsad, line 54: Would it be okay for you to auto-format all the tests in a follow-up PR?

We typically have slightly different code-styles in `kotlinx-` libraries, and it's okay (and it's quite pointless to establish *the* way across a bunch of independent libraries), but it would be way too easy to confuse
`@Test foo()` vs `@Test\nfoo()` and switch the context all the time. We prefer the latter

core/common/src/-CommonPlatform.kt
```diff
@@ -21,30 +21,18 @@
...
+public expect open class IOException(message: String?, cause: Throwable?) : Exception {
+  public constructor(message: String? = null)
 }
 
-expect class ProtocolException(message: String) : IOException
-
-expect open class EOFException(message: String? = null) : IOException
-
-expect class FileNotFoundException(message: String? = null) : IOException
+public expect open class EOFException(message: String? = null) : IOException
```
qwwdfsad, line 30: Does it make sense to make `EOFException` public right from the start?

core/common/src/Sink.kt
```diff
@@ -21,295 +21,139 @@
 package kotlinx.io
 
 /**
- * A sink that keeps a buffer internally so that callers can do small writes without a performance
- * penalty.
+ * A sink that facilitates typed data writes and keeps a buffer internally so that caller can write some data without
```
qwwdfsad, line 24: Here and in `Source` (and `Raw*` counterpart), it is probably worth explaining the following:

* Exception contract. Right now, it's a bit unclear what kind of exceptions are allowed from `read*` methods, it would be nice to streamline it a bit (probably it's a good idea to allow only `IOException` from the underlying sources and ISE/IAE for incorrect arguments/states). [An example](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-k-serializer/)
* Naming scheme for methods (https://github.com/Kotlin/kotlinx-io/issues/137#issuecomment-1594324577) explicitly spelled out is a great thing as well

core/common/src/SinkExt.kt
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * Writes four bytes containing [int], in the little-endian order, to this sink.
+ *
+ * @param int the integer to be written.
+ *
+ * @throws IllegalStateException when the sink is closed.
+ */
+public fun <T: Sink> T.writeIntLe(int: Int) {
+    this.writeInt(int.reverseBytes())
```
qwwdfsad, line 32: Also a note for a follow-up PR: it would be nice to reformat (but now it has to be a bit more accurate) the main sources as well to have 4 spaces as the default indent, not 2; the reasoning is as usual -- to have kotlinx consistent across the namespace and be the example

core/common/src/RawSource.kt
```diff
@@ -21,60 +21,31 @@
...
+   * Returns the number of bytes read, or -1 if this source is exhausted.
+   *
+   * @param sink the destination to write the data from this source.
+   * @param byteCount the number of bytes to read.
    *
-   * Note that it is always necessary to call [close] on a source, even if it has been canceled.
+   * @throws IllegalArgumentException when [byteCount] is negative.
+   * @throws IllegalStateException when the source is closed.
    */
-  fun cancel()
```
qwwdfsad, line 72: Worth creating an issue for this change to [re]evaluate it later

qwwdfsad (changes requested):
Tremendous work!

I've skimmed through the API, focusing mostly on the public signatures and their documentation, and tried to leave mostly generalized comments, to speed up the convergence of the change.
Minor things like code style, proof-reading and/or internal code structure we can always address later

core/common/src/SinkExt.kt
```diff
@@ -0,0 +1,240 @@
...
+    if (v < 0L) {
+        v = -v
+        if (v < 0L) { // Only true for Long.MIN_VALUE.
+            return writeUtf8("-9223372036854775808")
+        }
+        negative = true
+    }
+
+    // Binary search for character width which favors matching lower numbers.
+    var width =
```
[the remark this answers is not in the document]
fzhinkin: I was looking into https://johnnylee-sde.github.io/Fast-unsigned-integer-to-string/, but I have to sweep the dust from my Hacker's Delight copy first. :)

core/common/src/RawSource.kt
fzhinkin: I thought it should be added back with async support (although, there is no issue for that yet).

core/common/src/Buffer.kt
```diff
@@ -20,396 +20,124 @@
...
 
   /**
-   * Returns the number of bytes in segments that are not writable. This is the number of bytes that
-   * can be flushed immediately to an underlying sink without harming throughput.
+   * Returns the number of bytes in segments that are fully filled and are no longer writable.
+   *
+   * This is the number of bytes that can be flushed immediately to an underlying sink without harming throughput.
    */
-  fun completeSegmentByteCount(): Long
+  public fun completeSegmentByteCount(): Long
```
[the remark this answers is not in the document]
fzhinkin: And it should be internal.

core/jvm/test/JvmPlatformTest.kt
fzhinkin: Yes, I'll definitely be okay

core/common/src/Sink.kt
fzhinkin: I'll wait a few days to make sure there will be no suggestions to correct the latest version of the naming scheme from #137 and then fix the docs.

core/jvm/src/Buffer.kt
fzhinkin: @swankjesse Do you remember why it was implemented in the Okio in the first place (https://github.com/square/okio/commit/dad305d3c65647a9ea050f061a3544b0f4a67fed#diff-00883708f3cfa69f269509f3ade90c4e09e67bbd2a7a488e040cb82812284ba1R42)?

swankjesse: Being able to deep-copy a Buffer is nice and I’d keep that behavior.

As for implementing Cloneable, I very rarely see APIs that do anything with `Cloneable` so I recommend dropping it.

core/common/src/DelicateIoApi.kt
```diff
@@ -0,0 +1,15 @@
...
+package kotlinx.io
+
+@MustBeDocumented
+@Retention(AnnotationRetention.BINARY)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING,
+    message = "This is a delicate API and its use requires care. " +
+            "Make sure you fully read and understand documentation of the declaration that is marked as a delicate API."
+)
+public annotation class DelicateIoApi
```
swankjesse, line 15: Love this

core/common/src/SourceExt.kt
```diff
@@ -0,0 +1,309 @@
...
+ *
+ * @throws NumberFormatException if the found digits do not fit into a `long` or a decimal
+ * number was not present.
+ * @throws EOFException if the source is exhausted before a call of this method.
+ * @throws IllegalStateException when the source is closed.
+ */
+@OptIn(DelicateIoApi::class)
+public fun Source.readDecimalLong(): Long {
+    require(1)
+    var b = readByte()
```
swankjesse, line 56: This should consume zero bytes if it isn’t a decimal value

core/common/src/SourceExt.kt
```diff
@@ -0,0 +1,309 @@
...
+ *
+ * @throws NumberFormatException if the found hexadecimal does not fit into a `long` or
+ * hexadecimal was not found.
+ * @throws EOFException if the source is exhausted before a call of this method.
+ * @throws IllegalStateException when the source is closed.
+ */
+@OptIn(DelicateIoApi::class)
+public fun Source.readHexadecimalUnsignedLong(): Long {
+    require(1)
+    var b = readByte()
```
swankjesse, line 121: Similarly, this aught to consume zero bytes on bad input

core/common/src/SourceExt.kt
```diff
@@ -0,0 +1,309 @@
...
+@OptIn(DelicateIoApi::class)
+@Suppress("NOTHING_TO_INLINE")
+private inline fun Source.readByteArrayImpl(size: Long): ByteArray {
+    var arraySize = size
+    if (size == -1L) {
+        var fetchSize = Int.MAX_VALUE.toLong()
+        while (buffer.size < Int.MAX_VALUE && request(fetchSize)) {
+            fetchSize *= 2
+        }
+        if (buffer.size >= Int.MAX_VALUE) {
```
swankjesse, line 215: check(...)

core/jvm/src/Buffer.kt
swankjesse: In Okio two UTF-8 buffers with the same length, and length > 64 will have identical toString() values.

core/common/src/internal/-RealBufferedSource.kt
fzhinkin: I'll update it to `Source is closed.` / `Sink is closed.`. Not sure if there is anything else we can say to a user about the problem.

core/common/src/-CommonPlatform.kt
fzhinkin: We're declaring that some methods may throw `EOFException`, so I expect someone will try to catch it.

buildSrc/src/main/kotlin/Platforms.kt
```diff
@@ -38,20 +43,28 @@ private val appleTargets = listOf(
...
 
-val nativeTargets = appleTargets + linuxTargets + mingwTargets
+private val androidTargets = listOf(
+    "androidNativeArm32",
+    "androidNativeArm64",
+    "androidNativeX64",
+    "androidNativeX86"
+)
+
+val nativeTargets = appleTargets + linuxTargets + mingwTargets + androidTargets
```
whyoleg, line 67: minor: new kotlin target hierarchy API could be used instead: https://kotlinlang.org/docs/whatsnew1820.html#new-approach-to-source-set-hierarchy

core/Module.md
```diff
@@ -0,0 +1,79 @@
...
+and then back to an object using `kotlinx.io`. Please note that the example aimed to show `kotlinx-io` API in action,
+rather than to provide a robust BSON-serialization.
+```kotlin
+data class Message(val timestamp: Long, val text: String) {
+    companion object
+}
+
+fun Message.toBson(sink: Sink) {
+    val buffer = Buffer()
+    buffer.writeByte(0x9)                        // UTC-timestamp field
```
whyoleg, line 31: example should be updated after dropping returning `this` in `write` methods

core/common/src/SinkExt.kt
```diff
@@ -0,0 +1,272 @@
...
+}
+
+/**
+ * Writes am unsigned byte to this sink.
+ *
+ * @param byte the byte to be written.
+ *
+ * @throws IllegalStateException when the sink is closed.
+ */
+public fun Sink.writeByte(byte: UByte) {
```
whyoleg, line 184: just want to understand: why it's not `writeUByte`? (same for other unsigned writes)
now it's a little inconsistent with `read` part: `readUByte(): UByte` but `writeByte(value: UByte)`

core/common/src/SourceExt.kt
```diff
@@ -0,0 +1,318 @@
...
+}
+
+
+/**
+ * Removes exactly `sink.length` bytes from this source and copies them into [sink].
+ *
+ * @throws EOFException when the requested number of bytes cannot be read.
+ * @throws IllegalStateException when the source is closed.
+ */
+public fun Source.readTo(sink: ByteArray) {
```
whyoleg, line 236: is there a reason to not to have `start/end` indices in this declaration?

core/jvm/src/BufferExtJvm.kt
```diff
@@ -0,0 +1,204 @@
...
+  }
+}
+
+/**
+ * Writes up to [ByteBuffer.remaining] bytes from this buffer to the sink.
+ * Return the number of bytes written.
+ *
+ * @param sink the sink to write data to.
+ */
+public fun Buffer.readAtMostTo(sink: ByteBuffer): Int {
```
whyoleg, line 152: Im a bit confused with interop functions here, both naming and what should be provided.
we have `readAtMostTo(sink, byteCount)` and `readAtMost(array, start, end)` + we have `readTo(array)` (which will read to size) and others
May be it make sense to recheck on which functions (and their names) we should provide for interop (in separate issue?)
May be it's ok now, though, for now it's a harder to understand them comparing to `common` API

whyoleg (commented):
just some more or less minor questions on public API

core/jvm/src/BufferExtJvm.kt
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * Writes [byteCount] bytes from this buffer to [out].
+ *
+ * @param out the [OutputStream] to write to.
+ * @param byteCount the number of bytes to be written, [Buffer.size] by default.
+ *
+ * @throws IllegalArgumentException when [byteCount] is negative or exceeds the buffer size.
+ */
+public fun Buffer.writeTo(out: OutputStream, byteCount: Long = size) {
```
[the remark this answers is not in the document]
fzhinkin: It should, thanks!

core/jvm/src/JvmCore.kt
```diff
@@ -107,56 +103,62 @@ private open class InputStreamSource(
...
  */
-@Throws(IOException::class)
-fun Socket.sink(): RawSink {
-  return object : OutputStreamSink(getOutputStream()) {
-    override fun cancel() {
-      this@sink.close()
-    }
-  }
-}
+public fun Socket.sink(): RawSink = OutputStreamSink(getOutputStream())
```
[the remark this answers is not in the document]
fzhinkin: Not sure if I get the idea. Could you please elaborate?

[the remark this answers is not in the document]
whyoleg: why we need `Socket.sink` as we can easily do `socket.outputStream.sink()`. Previously in Okio there was different implementation, which overrides `cancel`, now there is no such functionality.
Of course if there will be later introduced Async IO it can be revisited, though, I believe there are other java NIO primitives suited for Async IO (debatable of course).
This is just thoughts on reducing public API regarding what is **really** needed regarding interop, so this can be left as is for now

core/common/src/SinkExt.kt
fzhinkin: I don't have a strong opinion here. It should probably follow the common naming convention and be `writeU<type>`, as you suggested.

core/common/src/SourceExt.kt
fzhinkin: No, there isn't. I'll add indices, thanks for pointing to it.

core/jvm/src/JvmCore.kt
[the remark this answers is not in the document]
fzhinkin: Got it, thanks. Extensions for `j.i.File` and `j.n.Path` could be also removed. I'll consider that, thanks for the suggestion!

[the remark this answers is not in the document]
fzhinkin: https://github.com/Kotlin/kotlinx-io/issues/143

whyoleg:
Question/Suggestion about naming conversion functions:
* `RawSource.buffer(): Source` - may be rename it to `buffered` - it's more aligned with [Kotlin naming conventions](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names) and also sounds more logical (same for `Sink`)
* `Source.channel(): ReadableByteChannel` - may be rename it to `asByteChannel` (or `toByteChannel`, though I think `as` is better) - idea is the same, to align more with Kotlin naming convention (same for `Sink` and `inputStream`/`outputStream`
* May be it's also make sense to introduce deprecated functions like `Source.buffer(): Source` (same for `Sink`) to show that it's doesn't make sense?

fzhinkin:
Regarding the performance: [results @ jmh.morethan.io](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fzhinkin/2709e9c45ae48903fc2f3cea18c0db2b/raw/f342c7f4834d1f245b3fbf9e6f8bf4d838870ce8/00-kotlinx-io-benchmarking-result-before-pr136.json,https://gist.githubusercontent.com/fzhinkin/2709e9c45ae48903fc2f3cea18c0db2b/raw/f342c7f4834d1f245b3fbf9e6f8bf4d838870ce8/01-kotlinx-io-benchmarking-result-after-pr136.json)

As anticipated, there's significant degradation compared to Okio in cases that use continuous scans within the buffer's content or write continuous byte sequences to the buffer. These issues will be addressed in https://github.com/Kotlin/kotlinx-io/issues/135.

core/build.gradle.kts
```diff
@@ -3,10 +3,13 @@
  * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENCE file.
  */
 
+import org.jetbrains.dokka.gradle.DokkaTaskPartial
 import org.jetbrains.kotlin.gradle.plugin.*
 
 plugins {
     kotlin("multiplatform")
+    id("org.jetbrains.kotlinx.kover") version "0.7.0"
```
shanshin, line 11: Latest version is `0.7.1`

core/Module.md
```diff
@@ -0,0 +1,84 @@
+# Module kotlinx-io-core
+
+The module provides core multiplatform IO primitives and integrates it with platform-specific APIs.
+
+`kotlinx-io` aims to provide a concise but powerful API along with efficient implementation.
```
shanshin, line 5: Maybe something like `kotlinx-io core` is better, because this is a file with a description of `kotlinx-io-core`, and here the text is written as if about the whole IO?

core/common/src/-Util.kt
```diff
@@ -19,17 +19,28 @@
...
 
-@SharedImmutable
 internal val HEX_DIGIT_CHARS =
   charArrayOf('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f')
 
 internal fun checkOffsetAndCount(size: Long, offset: Long, byteCount: Long) {
-  if (offset or byteCount < 0 || offset > size || size - offset < byteCount) {
-    throw ArrayIndexOutOfBoundsException("size=$size offset=$offset byteCount=$byteCount")
+  if (offset < 0 || offset > size || size - offset < byteCount || byteCount < 0) {
+    throw IllegalArgumentException("offset: $offset, byteCount: $byteCount, size: $size")
```
shanshin, line 31: According to this message, it is not entirely clear what exactly the error is, without knowing the implementation of the method.

Perhaps it is worth at least specifying in the message how these values should be related?

core/common/src/-Util.kt
```diff
@@ -19,17 +19,28 @@
...
+  }
+}
+
+internal inline fun checkBounds(size: Int, startIndex: Int, endIndex: Int) =
+  checkBounds(size.toLong(), startIndex.toLong(), endIndex.toLong())
+
+internal fun checkBounds(size: Long, startIndex: Long, endIndex: Long) {
+  if (startIndex < 0 || endIndex > size) {
+    throw IndexOutOfBoundsException("startIndex: $startIndex, endIndex: $endIndex, size: $size")
```
shanshin, line 40: Same as above

core/build.gradle.kts
fzhinkin: Thanks! Also, I'll update other dependencies.

core/common/src/-Util.kt
fzhinkin: Sure, I'll add more context to the message.

fzhinkin:
> Question/Suggestion about naming conversion functions:
> 
> * `RawSource.buffer(): Source` - may be rename it to `buffered` - it's more aligned with [Kotlin naming conventions](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names) and also sounds more logical (same for `Sink`)
> * `Source.channel(): ReadableByteChannel` - may be rename it to `asByteChannel` (or `toByteChannel`, though I think `as` is better) - idea is the same, to align more with Kotlin naming convention (same for `Sink` and `inputStream`/`outputStream`
> * May be it's also make sense to introduce deprecated functions like `Source.buffer(): Source` (same for `Sink`) to show that it's doesn't make sense?

Thanks for the suggestion!

I think it's better to return the buffered source itself from the `Source::buffered` / `Sink::buffered` instead of introducing a deprecated method.

core/common/src/Buffer.kt
```diff
@@ -20,396 +20,679 @@
...
-  fun copyTo(
+  override fun emit(): Unit = Unit
+
+  /**
+   * This method does not affect the buffer's content as there is no upstream to write data to.
+   */
+  override fun flush(): Unit = Unit
+
+  /**
+   * Copy bytes from this buffer's subrange starting at [startIndex] and ending at [endIndex], to [out] buffer.
```
shanshin, line 207: Does this operation not consume bytes from the current buffer?
If so, can it be specified explicitly?

core/common/src/Core.kt
```diff
@@ -20,34 +20,34 @@
...
-fun RawSink.buffer(): Sink = RealSink(this)
+public fun RawSink.buffer(): Sink = RealSink(this)
 
-/** Returns a sink that writes nowhere. */
-@JvmName("blackhole")
-fun blackholeSink(): RawSink = BlackholeSink()
+/**
+ * Returns a sink that writes nowhere.
+ */
+public fun blackholeSink(): RawSink = BlackholeSink()
```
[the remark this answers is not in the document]
fzhinkin: Speaking of `blackholeSink`, I can't find anything related in stdlib, but I like the naming from the Apache Flink: they have `DiscardingSink`. The name clearly translates what the sink does.

whyoleg:
> I think it's better to return the buffered source itself from the Source::buffered / Sink::buffered instead of introducing a deprecated method.

May be it will be sufficient to do something like this:
```kotlin
public fun RawSource.buffered(): Source = when(this) {
    is Source -> this
    else      -> RealSource(this)
}
```
(same for sink)

core/common/test/CommonPlatformTest.kt
```diff
@@ -24,24 +24,24 @@ package kotlinx.io
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
-class CommonOkioKotlinTest {
+class CommonPlatformTest {
   @Test fun sourceBuffer() {
-    val source = Buffer().writeUtf8("a")
-    val buffered = (source as RawSource).buffer()
+    val source = Buffer().also { it.writeUtf8("a") }
+    val buffered = (source as RawSource).buffered()
```
whyoleg, line 30: BTW: I believe, that this cast is not needed anymore. In Okio, there is deprecated function on `BufferedSource/BufferedSink` called [`buffer()`](https://github.com/square/okio/blob/81bce1a30af244550b0324597720e4799281da7b/okio/src/jvmMain/kotlin/okio/BufferedSink.kt#L30) to get underlying `Buffer`. in `kotlinx-io` we now have only `val buffer: Buffer`

fzhinkin: The cast is still required when someone is trying to invoke `buffered` on a `Buffer` because it's ambiguous whether `Sink::buffered` or `Source::buffered` should be called.

I'm wondering if someone (besides tests like this) would actually want to have multiple buffered sinks or sources on top of each other.

swankjesse: Yep, you got it. The only time you’d buffer a buffer is in a test.

swankjesse:
@whyoleg I recommendation against a feature to prevent multiple buffer wrappers. We were occasionally tempted to do this for Okio!

The runtime cost of redundant buffering is nearly zero. (This is a consequence of how data is moved and not copied between layers.)

And it creates quite surprising behaviour consequences. For example it makes code that forgets to call ‘emit()’ or ‘flush()’ work sometimes when it shouldn’t.

core/common/test/CommonBufferTest.kt
```diff
@@ -47,29 +47,26 @@ class CommonBufferTest {
...
-//        .write("74c999cb8872616ec999cb8c73c3b472".decodeHex())
-//        .toString()
-//    )
-//    assertEquals(
-//      "[hex=0000000000000000000000000000000000000000000000000000000000000000000000000000" +
-//        "0000000000000000000000000000000000000000000000000000]",
-//      Buffer().write(ByteArray(64)).toString()
-//    )
-//  }
+  @Test fun bufferToString() {
```
shanshin, line 50: We could also add a write/read test for a surrogate pair like https://github.com/Kotlin/kotlinx.serialization/blob/ba7d17a15329840a85e823c6d4bdc3595d8a19c6/formats/json-tests/commonTest/src/kotlinx/serialization/features/EmojiTest.kt#L12

shanshin: I see, this test in `Utf8Test`

fzhinkin:
> @whyoleg I recommendation against a feature to prevent multiple buffer wrappers. We were occasionally tempted to do this for Okio!
> 
> The runtime cost of redundant buffering is nearly zero. (This is a consequence of how data is moved and not copied between layers.)
> 
> And it creates quite surprising behaviour consequences. For example it makes code that forgets to call ‘emit()’ or ‘flush()’ work sometimes when it shouldn’t.

For me, it seems like both modes are legit - one may want to always wrap a `Source` into another `Source`, at least for testing, while in some cases, it may add an overhead (yeah, copying segments between buffers is fast, but the cost is non zero).
I think we can leave old behavior for now and discuss alternatives in a separate issue.

whyoleg:
@swankjesse after looking for a while on examples in kdoc of `flush` and `emit` in Okio - I'm now thinking, that you are right   that it will be not really expected behaviour if we will just return same object.

shanshin (commented):
Waiting for completion

core/jvm/src/BufferExtJvm.kt
fzhinkin: https://github.com/Kotlin/kotlinx-io/issues/143#issuecomment-1604232178

whyoleg:
@fzhinkin BTW, can update preview documentation (https://github.com/fzhinkin/kotlinx-io-dokka-docs-preview) after all latest changes - it's much more easier to check all available public API there (as members and extensions) are there in one list and sorted by name ?

buildSrc/src/main/kotlin/Platforms.kt
fzhinkin: https://github.com/Kotlin/kotlinx-io/issues/147

fzhinkin:
> @fzhinkin can you update preview documentation (https://github.com/fzhinkin/kotlinx-io-dokka-docs-preview) after all latest changes - it's much more easier to check all available public API there as members and extensions are there in one list and sorted by name ?

@whyoleg done, thanks for reminding me!

core/common/src/Buffer.kt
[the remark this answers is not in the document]
fzhinkin: So, I don't think we need to change the name as long as it's an internal API

buildSrc/src/main/kotlin/Platforms.kt
```diff
@@ -17,7 +17,6 @@ fun KotlinMultiplatformExtension.configureNativePlatforms() {
     tvosSimulatorArm64()
```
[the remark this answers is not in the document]
fzhinkin: Done

shanshin (approved):

core/common/test/AbstractSinkTest.kt
```diff
@@ -25,26 +25,60 @@ import kotlin.test.Test
...
 import kotlin.test.assertFailsWith
 
-class BufferSinkTest : AbstractBufferedSinkTest(BufferedSinkFactory.BUFFER)
-class RealBufferedSinkTest : AbstractBufferedSinkTest(BufferedSinkFactory.REAL_BUFFERED_SINK)
+class BufferSinkTest : AbstractSinkTest(SinkFactory.BUFFER)
+class RealSinkTest : AbstractSinkTest(SinkFactory.REAL_BUFFERED_SINK)
 
-abstract class AbstractBufferedSinkTest internal constructor(
-  factory: BufferedSinkFactory
+abstract class AbstractSinkTest internal constructor(
```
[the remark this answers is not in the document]
fzhinkin: I'm planning to reformat all the code in separate PR once this one will be merged (https://github.com/Kotlin/kotlinx-io/pull/136#discussion_r1232058811)

core/common/src/Utf8.kt
```diff
@@ -118,445 +117,415 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
-internal const val REPLACEMENT_CODE_POINT: Int = REPLACEMENT_CHARACTER.code
+/**
+ * Encodes [codePoint] in UTF-8 and writes it to this sink.
+ *
+ * @param codePoint the codePoint to be written.
+ *
+ * @throws IllegalStateException when the sink is closed.
+ */
+@OptIn(DelicateIoApi::class)
+public fun Sink.writeUtf8CodePoint(codePoint: Int): Unit =
```
qwwdfsad, line 128: I'd suggest hiding all code-point signatures as well.

We plan adding a bit more nuanced code point support to multiplatform (https://youtrack.jetbrains.com/issue/KT-30509), would be a shame to have all the cool names taken

core/common/src/RawSource.kt
qwwdfsad: It seemed like this method was useful even without explicit asynchrony (and was, in fact, not related to coroutines-based cancellation); I might be mistaken though, anyway worth initiating the discussion

qwwdfsad (approved):
Great job!

Left a few minor comments, feel free to address them as you see fit

core/common/src/Utf8.kt
fzhinkin: Done

core/common/src/Utf8.kt
```diff
@@ -66,26 +66,25 @@
...
+ * @throws IndexOutOfBoundsException when [startIndex] or [endIndex] is out of range of string indices.
+ * @throws IllegalArgumentException when `startIndex > endIndex`.
  */
-@JvmOverloads
-@JvmName("size")
-fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
-  require(beginIndex >= 0) { "beginIndex < 0: $beginIndex" }
-  require(endIndex >= beginIndex) { "endIndex < beginIndex: $endIndex < $beginIndex" }
-  require(endIndex <= length) { "endIndex > string.length: $endIndex > $length" }
+public fun String.utf8Size(startIndex: Int = 0, endIndex: Int = length): Long {
```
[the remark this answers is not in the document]
fzhinkin: Done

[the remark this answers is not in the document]
joffrey-bion: I was actually relying on this function in Krossbow, because I need to truncate a string based on its length in UTF-8 bytes (not characters). The use case is the "close reason" (the text payload) of the web socket `Close` frame. As per the [specification](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5), the payload of all control frames (Close, Ping, Pong) must be <= 125 bytes.

Without this function, I need to encode the string into a byte array, then check if it's too long, and then truncate. With the `utf8Size` function, I could optimize the happy path by not copying bytes around if the string was short enough.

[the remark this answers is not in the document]
joffrey-bion: I opened an issue for this https://github.com/Kotlin/kotlinx-io/issues/258. I figured it's a better place than here. Sorry for the noise

core/common/src/Utf8.kt
sandwwraith: `writeUtf8CodePoint` is actively used in Json integration (to do e.g. `write(":")`), so I suggest making it public now.

Issue #133, opened 2023-06-12
Implement ByteStrings


fzhinkin:
Immutable array wrappers are ubiquitous in JVM ecosystem. These wrappers usually provide different handy methods like content-based equals/hashCode (allowing to use of such wrappers in hash-based data structures), string-like APIs to search for particular values or subsequences within a wrapper, copy or create a subsequence, and so on. Also, these wrappers guarantee that the backing array won’t change, no matter what. The latter allows to overcome the need for defensive array copies.

Okio already has ByteStrings, but these are intermingled with Okio internals. It would be nice to provide an implementation based on Okio ByteStrings, supported by `kotlinx-io`, but shipped as a different artifact so that users could include it in their projects separately.

swankjesse:
Making Okio’s ByteString implement Comparable has proven quite handy in practice too.

For example, to deterministically order a bunch of protobufs I can sort by their encodings. Or perhaps I sort their hashes!

fzhinkin:
Related issue: https://github.com/Kotlin/kotlinx-io/issues/115

Issue #142, opened 2023-06-16
Break the curse of buildSrc and migrate to a more conventional build configuration


qwwdfsad:
...while it's not too late.

`buildSrc` is rather a legacy from the early days of KTS adoption that we kept piggybacking on. It's accompanied by a steady yet manageable stream of minor issues, incompatibilities and general discomfort.

Would be nice to use composite builds (via `includeBuild`) instead

shanshin:
Example of composite build in Kover https://github.com/Kotlin/kotlinx-kover/tree/main/build-logic. 
Also in Dokka https://github.com/Kotlin/dokka/tree/master/build-logic and [article](https://proandroiddev.com/stop-using-gradle-buildsrc-use-composite-builds-instead-3c38ac7a2ab3).

EndzeitBegins:
Is using `buildSrc` still a problem? I was under the impression that with the newest Gradle versions `buildSrc` would behave mostly like an included build, e.g. [in of the Gradle docs](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) it's said that:

> [...] The directory buildSrc is treated as an included build. [...]

In the [release notes of Gradle 8.x](https://gradle.org/whats-new/gradle-8/#build-configuration) it is also stated that:

> [...] At scale, sometimes included builds are more performant, and sometimes buildSrc are. The performance depends on how exactly things are structured. [...]
> In working towards the unification of buildSrc and included builds, buildSrc is changing to behave more like included builds. [...]
> If you’re using buildSrc already, keep using it and enjoy the new features. If you’re using included builds already, keep using them. [...]

Issue #134, opened 2023-06-12
Provide samples showing how to use kotlinx-io API


fzhinkin:
Currently, there are no samples (besides unit tests) showing how to use `kotilnx-io`. This should be fixed to allow users quickly grasp the API and start using it.

Some samples could be taken from Okio.

PR #148, opened 2023-06-23
ByteString implementation

fzhinkin:
Implemented ByteString - an immutable wrapper around byte sequence and added base support to the core IO library.

Missing stuff:
- [x] Module documentation

Closes https://github.com/Kotlin/kotlinx-io/issues/133


bytestring/src/commonMain/kotlin/UnsafeByteStringOperations.kt
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Collection of helper functions providing unsafe access to the [ByteString]'s underlying byte sequence or allowing
+ * to wrap byte arrays into [ByteString] without copying the array.
+ *
+ * These functions are provided for performance sensitive cases where it is known that the data accessed
+ * in an unsafe manner won't be modified. Modification of the data backing byte strings may lead to unpredicted
+ * consequences in the code using the byte string and should be avoided at all costs.
+ */
+@UnsafeByteStringApi
+public class UnsafeByteStringOperations {
```
qwwdfsad, line 19: you can use `public object UnsafeByteStringOperations` instead

bytestring/src/commonMain/kotlin/ByteStringBuilder.kt
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public val capacity: Int
+        get() = buffer.size
+
+    /**
+     * Returns a new [ByteString] wrapping all bytes written to this builder.
+     *
+     * There will be no additional allocations or copying of data when `size == capacity`.
+     */
+    public fun toByteString(): ByteString {
```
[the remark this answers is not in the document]
fzhinkin: > does it make sense to reset current buffer? 

`StringBuilder::toString` don't reset the builder, so I decided to abstain from doing it for `ByteStringBuilder`.

> Or user will be able to append bytes after calling toByteString and so mutate ByteString data if buffer.size == size.

Any attempt to append something to the builder when `buffer.size == size` will end up with the backing array's extension (so a new array will be allocated and data from the old one, potentially wrapped by a byte string, will be copied into it), so the old value won't be affected.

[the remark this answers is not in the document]
fzhinkin: > BTW, StringBuilder will always create a copy when building String

The main motivation for having this builder is to have a way to create a ByteString of known size without doing extra copies (Yes, there is `unsafeWrap`, but it's *unsafe*, while the builder provides a safe alternative).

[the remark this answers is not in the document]
whyoleg: yeah, never mind, somehow I missed part, that we have only `append` methods, and not full `StringBuilder` like API with `replace`/`insert`/`delete`/etc...
so just ignore this comments :)

core/common/src/ByteStringExt.kt
```diff
@@ -0,0 +1,117 @@
...
+ * @param byteString the byte string whose subsequence should be written to a sink.
+ * @param startIndex the first index (inclusive) to copy data from the [byteString].
+ * @param endIndex the last index (exclusive) to copy data from the [byteString]
+ *
+ * @throws IndexOutOfBoundsException when [startIndex] or [endIndex] is out of range of [byteString] indices.
+ * @throws IllegalArgumentException when `startIndex > endIndex`.
+ * @throws IllegalStateException if the sink is closed.
+ */
+@OptIn(DelicateIoApi::class)
+public fun Sink.write(byteString: ByteString, startIndex: Int = 0, endIndex: Int = byteString.size) {
```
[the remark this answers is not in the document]
fzhinkin: I can easily imagine how someone can use `readTo(ByteStringBuilder)`, but not sure how practical it would be.
So I would abstain from adding such extensions until there will be some convincing use cases.

bytestring/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
+import org.jetbrains.dokka.gradle.DokkaTaskPartial
+import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
+
+plugins {
+    kotlin("multiplatform")
+    id("org.jetbrains.kotlinx.kover") version "0.7.0"
+    id("org.jetbrains.dokka") version "1.8.10"
```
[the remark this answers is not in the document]
fzhinkin: Will do it along w/ #142. For now, just bumped up versions manually

bytestring/src/jvmMain/kotlin/ByteStringJvmExt.kt
```diff
@@ -0,0 +1,24 @@
...
+package kotlinx.io.bytestring
+
+import java.nio.charset.Charset
+
+/**
+ * Decodes the content of a byte string to a string using given [charset].
+ *
+ * @param charset the charset to decode data into a string.
+ */
+public fun ByteString.toString(charset: Charset): String = getBackingArrayReference().toString(charset)
```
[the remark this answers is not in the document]
fzhinkin: `ByteArray::toString` does the same thing. Not saying I'm a big fan of it, but it looks consistent with existing APIs.

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,452 @@
...
+     * These methods return reference to the underlying array, not to its copy.
+     * Consider using [toByteArray] if it's impossible to guarantee that the array won't be modified.
+     */
+    internal fun getBackingArrayReference(): ByteArray = data
+}
+
+/**
+ * Returns the range of valid byte indices for this byte string.
+ */
+public fun ByteString.indices(): IntRange = 0 until size
```
qwwdfsad, line 231: `indicies` rather should be a property than a method, as it is in other array-like structures

bytestring/common/src/ByteStringBuilder.kt
```diff
@@ -0,0 +1,131 @@
...
+ */
+public inline fun buildByteString(builderAction: ByteStringBuilder.() -> Unit): ByteString {
+    return ByteStringBuilder().apply(builderAction).toByteString()
+}
+
+/**
+ * Builds new byte string by populating newly created [ByteStringBuilder] initialized with the given [capacity]
+ * using provided [builderAction] and then converting it to [ByteString].
+ */
+public inline fun buildByteString(capacity: Int, builderAction: ByteStringBuilder.() -> Unit): ByteString {
```
qwwdfsad, line 129: Shouldn't it be `capacity = 0` to avoid an extra overload?

bytestring/common/src/unsafe/UnsafeByteStringOperations.kt
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public fun wrapUnsafe(array: ByteArray): ByteString = ByteString.wrap(array)
+
+    /**
+     * Returns a reference to the underlying array.
+     *
+     * These methods return reference to the underlying array, not to its copy.
+     * Consider using [ByteString.toByteArray] if it's impossible to guarantee that the array won't be modified.
+     */
+    public fun getByteArrayUnsafe(byteString: ByteString): ByteArray = byteString.getBackingArrayReference()
```
qwwdfsad, line 34: WDYT about explicitly confining lexical scope of the array, e.g. by 
```
.withByteArrayUnsafe { array -> ... }
```
instead of returning plain array?

bytestring/jvm/src/ByteStringJvmExt.kt
```diff
@@ -0,0 +1,24 @@
...
+package kotlinx.io.bytestring
+
+import java.nio.charset.Charset
+
+/**
+ * Decodes the content of a byte string to a string using given [charset].
+ *
+ * @param charset the charset to decode data into a string.
+ */
+public fun ByteString.toString(charset: Charset): String = getBackingArrayReference().toString(charset)
```
qwwdfsad, line 15: This one breaks the general rules of methods overloading: two overloads _usually_ differ in their input set, but have the same semantics.

This is not the case for `toString()` vs `toString(Charset)`

qwwdfsad (changes requested):

bytestring/jvm/src/ByteStringJvmExt.kt
fzhinkin: @whyoleg @qwwdfsad 
Maybe `encodeToString` / `decodeFromString` then? And `encodeToUtf8String` / `decodeFromUtf8String` for methods from the common source set.

whyoleg: I think, that we should use `decode` for conversion to `String` (same as in stdlib) as we a decoding bytes into characters, nor otherwise. In that case for `ByteString` creation it will be `ByteString.Companion.encodeFromString(String): ByteString`, and I don't really like the naming of `encode from`.

I do like stdlib-like naming more: `ByteString.decodeToString(): String` and `String.encodeToByteString(): ByteString` for UTF8 + overload with `Charset` on JVM (with possibility of multiplatform charsets in future) + may be overloads with indices (like we have in stdlib)
Though not sure, that adding extension to `String` is the best idea.

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,453 @@
...
+    return getBackingArrayReference().decodeToString()
+}
+
+/**
+ * Encodes a string into a byte sequence using UTF8-encoding and wraps it into a byte string.
+ *
+ * @param string the string to be encoded.
+ */
+public fun ByteString.Companion.fromUtf8String(string: String): ByteString {
+    return wrap(string.encodeToByteArray())
```
fzhinkin, line 443: `encodeToByteArray` behaves differently on JVM and Native when encoding invalid code points.

It may be fine, but primitives from the code module behave similarly on all targets (and the behavior is the same as `encodeToByteArray`'s on JVM).

So we have to either hoist UTF-8 support from the code module into a separate module and reuse it here or change the behavior in the core module to match stdlib's behavior.

Either way, I'd rather postpone a fix and discuss it separately.

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,456 @@
...
+     * @throws IllegalArgumentException when `startIndex > endIndex`.
+     */
+    public fun substring(startIndex: Int, endIndex: Int = size): ByteString = if (startIndex == endIndex) {
+        EMPTY
+    } else {
+        ByteString(data, startIndex, endIndex)
+    }
+
+    /**
+     * Compares a byte sequence wrapped by this byte string to a byte sequence wrapped by [other]
```
shanshin, line 174: It is worth adding either a more detailed description or examples of exactly how the comparison works.

If I am not mistaken, the lexicographic order does not strictly define the relation of values such as `(1, 2, 3) ? (1, 3)`

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,456 @@
...
+ * starting from the specified [startIndex].
+ *
+ * Behavior of this method is compatible with [CharSequence.indexOf].
+ *
+ * @param byte the value to search for.
+ * @param startIndex the index (inclusive) starting from which the [byte] should be searched.
+ */
+public fun ByteString.indexOf(byte: Byte, startIndex: Int = 0): Int {
+    for (i in max(startIndex, 0) until size) {
+        if (this[i] == byte) {
```
shanshin, line 248: Here and below: can't we use backing array to avoid unnecessary range checks?

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,456 @@
...
+ *
+ * Behavior of this method is compatible with [CharSequence.indexOf].
+ *
+ * @param byteString the value to search for.
+ * @param startIndex the index (inclusive) starting from which the [byteString] should be searched.
+ */
+public fun ByteString.indexOf(byteString: ByteString, startIndex: Int = 0): Int {
+    if (byteString.isEmpty()) return max(min(startIndex, size), 0)
+    for (i in max(startIndex, 0)..size - byteString.size) {
+        if (this[i] == byteString[0] && rangeEquals(i, byteString)) {
```
shanshin, line 267: Same about backing array and moving `byteString[0]` to var

bytestring/common/src/ByteString.kt
fzhinkin: I'll mention that the behavior is similar to `String::compareTo` and then add a sample comparing different byte strings in a separate PR (for https://github.com/Kotlin/kotlinx-io/issues/134)

bytestring/jvm/src/ByteStringJvmExt.kt
fzhinkin: Adopted the following naming:
- `ByteString.decodeToString(): String`
- `String.encodeToByteString(): ByteString`

whyoleg: nice!
last minor thing: as we not use `utf8` in naming anymore here, may be it make sense to do the same renaming in `core` module for `readUtf8` and `writeUtf8` (with `readString(Charset)`/`writeString(Charset)` on JVM) and replace common functions naming to use `readString`/`writeString`?

fzhinkin: Sure, sounds reasonable. Thanks for the suggestion! I filed https://github.com/Kotlin/kotlinx-io/issues/155 for it.

bytestring/common/src/ByteString.kt
whyoleg: Note: K/JS behaviour of `encodeToByteArray` is the same as in K/N (not sure about K/WASM). In that case only K/JVM behaviour is differs from other platforms.
And now in `kotlinx-io` we adopting K/JVM behaviour.
While it could be fine, it's a little inconsistent

core/common/src/BufferExt.kt
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.io
+
+import kotlinx.io.bytestring.ByteString
+import kotlinx.io.bytestring.buildByteString
+
```
qwwdfsad, line 10: Missed it during the original review.
We typically prefer extensions for `class Foo` either placed in the same file (-> class`FooKt`) for small and compact additions, or in plural form of the entity -- `Foos` (`Channels.kt`, `Strings.kt`, `Serializers.kt`). Bit nicer stacktraces, a bit more consistent

bytestring/build.gradle.kts
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+tasks.withType<DokkaTaskPartial>().configureEach {
+    dokkaSourceSets.configureEach {
+        includes.from("Module.md")
+
+        perPackageOption {
+            suppress.set(true)
+            matchingRegex.set(".*unsafe.*")
```
qwwdfsad, line 77: Do you think it's worth excluding it from the official documentation?

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,467 @@
...
+ * @param bytes a sequence of bytes to be wrapped.
+ */
+public fun ByteString(vararg bytes: Byte): ByteString = if (bytes.isEmpty()) {
+    ByteString.EMPTY
+} else {
+    ByteString.wrap(bytes)
+}
+
+/**
+ * An immutable wrapper around a byte sequence providing [String] like functionality.
```
qwwdfsad, line 41: It would be nice to document it extensively (how it can be used, what for and what it is semantically) in the follow-up PR

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,467 @@
...
+     * Wraps a copy of [data] subarray starting at [startIndex] and ending at [endIndex] into a byte string.
+     *
+     * @param data the array whose subarray should be copied and wrapped into a byte string.
+     * @param startIndex the start index (inclusive) of a subarray to copy, `0` by default.
+     * @param endIndex the end index (exclusive) of a subarray to copy, `data.size` be default.
+     *
+     * @throws IndexOutOfBoundsException when [startIndex] or [endIndex] is out of range of [data] array indices.
+     * @throws IllegalArgumentException when `startIndex > endIndex`.
+     */
+    public constructor(data: ByteArray, startIndex: Int = 0, endIndex: Int = data.size) :
```
qwwdfsad, line 74: Code style: please move the ctor above of all the properties

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,467 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kotlinx.io.bytestring
+
+import kotlin.math.max
+import kotlin.math.min
+
+private val HEX_DIGITS = "0123456789ABCDEF".toCharArray()
```
qwwdfsad, line 27: It's possible to place it in companion and save an extra classfile if you see it fit

bytestring/common/src/ByteString.kt
```diff
@@ -0,0 +1,467 @@
...
+        if (localData[i] == byte) {
+            return i
+        }
+    }
+    return -1
+}
+
+/**
+ * Returns the index within this char sequence of the last occurrence of the specified [byteString],
+ * starting from the specified [startIndex].
```
qwwdfsad, line 323: here and below worth explicitly mentioning `-1` as "not found"

bytestring/common/src/unsafe/UnsafeByteStringOperations.kt
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public fun wrapUnsafe(array: ByteArray): ByteString = ByteString.wrap(array)
+
+    /**
+     * Applies [block] to a reference to the underlying array.
+     *
+     * This method invokes [block] on a reference to the underlying array, not to its copy.
+     * Consider using [ByteString.toByteArray] if it's impossible to guarantee that the array won't be modified.
+     */
+    public fun withByteArrayUnsafe(byteString: ByteString, block: (ByteArray) -> Unit) {
```
qwwdfsad, line 34: Should be inline, otherwise it's an allocation in the performance-sensitive method

qwwdfsad (changes requested):

core/common/src/BufferExt.kt
qwwdfsad: We also use `@file:JvmMultifileClass` for files spread across platforms, but maybe it's a bit too redundant

fzhinkin: Thanks for fixing it!

bytestring/build.gradle.kts
fzhinkin: Otherwise, it looks like a Forbidden fruit - we're asking not to use it by any means and then making it visible to everyone.

bytestring/common/src/ByteString.kt
fzhinkin: https://github.com/Kotlin/kotlinx-io/pull/158

Issue #155, opened 2023-06-28
Unify ByteString/String conversions naming and read/write String methods naming from the core module


fzhinkin:
Currently, we use `ByteString.decodeToString` and `String.encodeToByteString` as names for conversion methods between `String` and `ByteString`, where non-parameterized functions convert to/from UTF-8, and JVM-specific extensions accept `Charset`.
At the same time, the core module uses different naming for methods reading/writing UTF-8 string and methods reading/writing strings using specific `Charset`:
- `Source.readUtf8`/`Sink.writeUtf8` to work with UTF-8 strings
- `Source.readString`/`Sink.writeString` to work with strings using the given `Charset` on JVM.

The naming is inconsistent and it seems reasonable to unify read/write methods naming with what we have for `ByteString`.
We can use `readString`/`writeString` w/o charset for UTF-8 strings (as these are treated as default in the library) and same-titled JVM-specific extensions accepting a `Charset`.

PR #156, opened 2023-06-28
Rename string read/writeUtf8 methods

fzhinkin:
Closes https://github.com/Kotlin/kotlinx-io/issues/155; see the ticket for the rationale.

### DIFF
--- a/benchmarks/src/commonMain/kotlin/BufferOps.kt
+++ b/benchmarks/src/commonMain/kotlin/BufferOps.kt
@@ -214,8 +214,8 @@ open class Utf8StringBenchmark : BufferRWBenchmarkBase() {
     @Benchmark
     fun benchmark(): String {
         val s = buffer.size
-        buffer.writeUtf8(string)
-        return buffer.readUtf8(buffer.size - s)
+        buffer.writeString(string)
+        return buffer.readString(buffer.size - s)
     }
 }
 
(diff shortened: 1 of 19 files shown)


whyoleg:
Question regarding `Charset` parameter position in functions. We have now:
* `Sink.writeString(string: String, charset: Charset, startIndex: Int = 0, endIndex: Int = string.length)`
* `Source.readString(charset: Charset): String`
* `Source.readString(byteCount: Long, charset: Charset): String`
* `ByteString.decodeToString(charset: Charset): String`
* `String.encodeToByteString(charset: Charset): ByteString`

What is better:
* what we have now, so `Charset` is last required argument (in case of `writeString`), and so f.e. can be even made optional (with default=Utf8) in future if multiplatform `Charset` will be supported
* put `Charset` as first parameter in all functions? (as in `writeString` and `readString(byteCount)`

Im fine with both variants, just wanted to ask :)

fzhinkin:
> What is better:
> 
> * what we have now, so `Charset` is last required argument (in case of `writeString`), and so f.e. can be even made optional (with default=Utf8) in future if multiplatform `Charset` will be supported

I didn't make it optional because there was already a function for writing UTF8 strings. And now, when it shares the name with Charset-based functions, we can encourage users to use a common function for UTF-8 strings instead of a JVM-specific one by requiring explicit charset value in the latter.

Possible future support for our own charsets in IO library is a great argument for supplying default value for the `charset` parameter, but charsets-support is not even on the horizon ATM, so I don't think it is worth preparing for it right now.

> * put `Charset` as first parameter in all functions? (as in `writeString` and `readString(byteCount)`

It looks strange to have a charset as a first parameter there. If we writing a string, then the string should be the first parameter and all other options controlling how it'll be written should follow it, not proceed.

qwwdfsad (approved):

shanshin (approved):

PR #161, opened 2023-07-03
Updated README w/ information about modules and KMP builds

fzhinkin:
Updated README w/ information about modules and KMP builds:

- included information about modules
- added badge for KDoc
- described how to add the dependency to KMP project

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,21 +32,21 @@ There are two `kotlinx-io` modules:
 ### Gradle
 
 Make sure that you have `mavenCentral()` in the list of repositories:
-```
+```kotlin
 repositories {
     mavenCentral()
 }
 ```
 
 Add the library to dependencies:
-```
+```kotlin
 dependencies {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,7 +41,20 @@ repositories {
 Add the library to dependencies:
 ```
 dependencies {
-    implementation("org.jetbrains.kotlinx:kotlinx-io:0.2.0-alpha-dev1")
+    implementation("org.jetbrains.kotlinx:kotlinx-io-core:0.2.0")
+}
+```
+
+In multiplatform projects, add a dependency to the `commonMain` source set dependencies:
+```
```
shanshin, line 49: For Gradle scripts we may specify 
`"```kotlin  "`

shanshin (approved):

README.md
fzhinkin: Fixed, thanks!

PR #174, opened 2023-07-12
Add Source and Sink extensions for Apple's NSInputStream and NSOutputStream

jeffdgr8:
Modeled after JVM's `InputStream.asSource()`, `Source.asInputStream()`, `OutputStream.asSink()`, and `Sink.asOutputStream()`, add extension functions to similarly map to and from Apple's `NSInputStream` and `NSOutputStream`.

I previously submitted [this PR](https://github.com/square/okio/pull/1123) to Okio. I've been using these extensions, along with my Okio fork, in my KMP library, pending availability in a public release.


fzhinkin:
@jeffdgr8 thank you for the PR!

Unfortunately, I forgot to mention in the contribution guideline that the PR should be created against the `develop` branch, so I switched the base here. It shouldn't be a problem as `master` and `develop` are almost identical right now.

whyoleg:
may be it make sense to also add the same functionality for [NSOutputStream](https://developer.apple.com/documentation/foundation/nsoutputstream) in one go?

core/apple/src/AppleCore.kt
```diff
@@ -0,0 +1,56 @@
...
+ * @sample kotlinx.io.samples.KotlinxIoSamplesApple.inputStreamAsSource
+ */
+public fun NSInputStream.asSource(): RawSource = NSInputStreamSource(this)
+
+private open class NSInputStreamSource(
+    private val input: NSInputStream,
+) : RawSource {
+
+    init {
+        input.open()
```
whyoleg, line 26: what will be if `NSInputStream` is already opened or closed? (before converting to source)

core/apple/src/AppleCore.kt
```diff
@@ -0,0 +1,56 @@
...
+            return -1
+        }
+        tail.limit += bytesRead.toInt()
+        sink.size += bytesRead
+        return bytesRead
+    }
+
+    override fun close() = input.close()
+
+    override fun toString() = "source($input)"
```
whyoleg, line 55: likely should be something like `RawSource($input)`

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,98 @@
...
+                    buffer?.pointed?.value = it.addressOf(s.pos).reinterpret()
+                    length?.pointed?.value = (s.limit - s.pos).convert()
+                    return true
+                }
+            }
+        }
+        return false
+    }
+
+    override fun hasBytesAvailable() = source.buffer.size > 0
```
whyoleg, line 71: could be replaced with `!source.exhausted()`

whyoleg (commented):
just minor comments

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    private val isClosed: () -> Boolean = when (source) {
+        is RealSource -> source::closed
+        is Buffer -> {
+            { false }
+        }
+    }
+
+    private var error: NSError? = null
+    private var pinnedBuffer: Pinned<ByteArray>? = null
```
whyoleg, line 34: is it possible not to store pinned value here? If stream will not be closed, it will likely be a memory leak of pinned instance

jeffdgr8:
> may be it make sense to also add the same functionality for [NSOutputStream](https://developer.apple.com/documentation/foundation/nsoutputstream) in one go?

I'm actually looking into implementing the `NSOutputStream` extensions for `Sink` as well. I don't need these in my own project, so haven't had a need for them. I can create a separate PR, or combine with this one.

core/apple/src/AppleCore.kt
jeffdgr8: `open()` should be idempotent, at least in the default `NSInputStream` implementations. There's no way to check the open/closed state of an `NSInputStream` other than getting an error if using it when closed.

core/apple/src/SourcesApple.kt
jeffdgr8: This is a solution I came up with with @swankjesse [in this discussion](https://github.com/square/okio/pull/1123#discussion_r928344672) on the Okio PR. Not pinning the buffer has the potential for the memory address returned to the caller in the `buffer` parameter to become invalid if GC decides to move it. Since leaving a stream open would likely leak other resources as well, I think this is the best we can do with the Kotlin/Objective-C memory interop as it is, unless there are other ideas?

core/apple/src/AppleCore.kt
whyoleg: Got it!

But I have one more minor: is it ok to open stream, even if it will be not used afterwards for some reason?
May be we should open it on first read?

Im not entirely know, how it used in iOS, so my questions can be wrong 😀 
Just trying to understand which behaviour is desirable here

core/apple/src/AppleCore.kt
whyoleg: I've already opened #177 for alignment of other contents

core/apple/src/AppleCore.kt
jeffdgr8: Actually, I take that last part back. (It's JVM's `InputStream` you can't get the open/closed status of.) There is [`NSStreamStatus`](https://developer.apple.com/documentation/foundation/nsstreamstatus).

jeffdgr8: I can also implement the status for `asNSInputStream()`. The question would be, should a `SourceNSInputStream` require calling `open()`, even though it's a no-op and unnecessary, or rather the status be `NSStreamStatusOpen` by default?

Mirroring the behavior of `[NSInputStream initWithData:]`, it is not open by default and reads fail until the stream is opened. So probably best to behave this way as well.

jeffdgr8:
I went ahead and added `NSOutputStream.asSink()` and `Sink.asNSOutputStream()` extensions as well.

core/apple/src/AppleCore.kt
```diff
@@ -0,0 +1,108 @@
...
+ */
+public fun NSOutputStream.asSink(): RawSink = OutputStreamSink(this)
+
+private open class OutputStreamSink(
+    private val out: NSOutputStream,
+) : RawSink {
+
+    @OptIn(UnsafeNumber::class)
+    override fun write(source: Buffer, byteCount: Long) {
+        if (out.streamStatus == NSStreamStatusNotOpen) out.open()
```
whyoleg, line 29: may be we also need to check for closed state here and throw right here?

core/apple/src/AppleCore.kt
```diff
@@ -0,0 +1,108 @@
...
+        var remaining = byteCount
+        while (remaining > 0) {
+            val head = source.head!!
+            val toCopy = minOf(remaining, head.limit - head.pos).toInt()
+            val bytesWritten = head.data.usePinned {
+                val bytes = it.addressOf(head.pos).reinterpret<UInt8Var>()
+                out.write(bytes, toCopy.convert()).toLong()
+            }
+
+            if (bytesWritten < 0L) throw IOException(out.streamError?.localizedDescription ?: "Unknown error")
```
whyoleg, line 41: Im not sure, but may be should use just `description` and not `localizedDescription` - I don't have preferences, and don't really know, how it works on JVM in such cases

core/apple/src/AppleCore.kt
```diff
@@ -0,0 +1,108 @@
...
+ */
+public fun NSInputStream.asSource(): RawSource = NSInputStreamSource(this)
+
+private open class NSInputStreamSource(
+    private val input: NSInputStream,
+) : RawSource {
+
+    @OptIn(UnsafeNumber::class)
+    override fun readAtMostTo(sink: Buffer, byteCount: Long): Long {
+        if (input.streamStatus == NSStreamStatusNotOpen) input.open()
```
whyoleg, line 79: same here for checking closed state

core/apple/src/AppleCore.kt
```diff
@@ -0,0 +1,108 @@
...
+        checkByteCount(byteCount)
+
+        val tail = sink.writableSegment(1)
+        val maxToCopy = minOf(byteCount, Segment.SIZE - tail.limit)
+        val bytesRead = tail.data.usePinned {
+            val bytes = it.addressOf(tail.limit).reinterpret<UInt8Var>()
+            input.read(bytes, maxToCopy.convert()).toLong()
+        }
+
+        if (bytesRead < 0L) throw IOException(input.streamError?.localizedDescription ?: "Unknown error")
```
whyoleg, line 91: same for `localizedDescription`

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,101 @@
...
+            }
+            status = NSStreamStatusOpen
+            maxLength.convert()
+        } catch (e: Exception) {
+            error = e.toNSError()
+            -1
+        }
+    }
+
+    private fun Buffer.writeNative(source: CPointer<uint8_tVar>?, maxLength: Int) {
```
whyoleg, line 76: minor: I would call it something like `copyFrom` - to better reflect what we do here
Or even put `Buffer` from receiver to ordinary parameter

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,101 @@
...
+                memcpy(it.addressOf(tail.pos), source + currentOffset, toCopy.convert())
+            }
+
+            currentOffset += toCopy
+            tail.limit += toCopy
+        }
+        size += maxLength
+    }
+
+    override fun hasSpaceAvailable() = true
```
whyoleg, line 92: may be we can return `!isClosed` here?

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    @OptIn(InternalIoApi::class)
+    override fun propertyForKey(key: NSStreamPropertyKey): Any? = when (key) {
+        NSStreamDataWrittenToMemoryStreamKey -> sink.buffer.readByteArray().toNSData()
```
whyoleg, line 96: `readByteArray` here will remove data from `buffer` - is it expected behaviour here?

core/apple/src/SourcesApple.kt
whyoleg: Ok, I see, that we can't do anything here to remove pinning.
Though:
1. as stated in documentation, this value is only valid until next read operation - so may be we can reset it on read, if available
2. may be we need to attach `Cleaner` for it? though not sure about overhead and is it really needed or not

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,121 @@
...
+                    buffer?.pointed?.value = it.addressOf(s.pos).reinterpret()
+                    length?.pointed?.value = (s.limit - s.pos).convert()
+                    return true
+                }
+            }
+        }
+        return false
+    }
+
+    private fun Buffer.readNative(sink: CPointer<uint8_tVar>?, maxLength: Int): Int {
```
whyoleg, line 98: Should we iterate over segments here while we not yet read `maxLength`?

whyoleg (commented):
a little more comments/questions :)

core/apple/test/NSOutputStreamSinkTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlinx.cinterop.pointed
+import kotlinx.cinterop.reinterpret
+import kotlinx.cinterop.value
+import platform.Foundation.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class NSOutputStreamSinkTest {
+    @Test
+    fun nsOutputStreamSink() {
```
fzhinkin, line 13: The test failed in CI w/ 
```
kotlin.AssertionError: Expected <97>, actual <-2145945247>.
```

core/apple/src/AppleCore.kt
jeffdgr8: [`NSError.localizedDescription`](https://developer.apple.com/documentation/foundation/nserror/1414418-localizeddescription?language=objc) is similar to [`Throwable.message`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-throwable/message.html) on the JVM, where [`NSObject.description`](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418746-description?language=objc) is the equivalent of [`Any.toString()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/to-string.html). I actually remember this being confusing myself when I was first learning Objective-C for iOS, having more experience with Java a decade ago.

core/apple/src/AppleCore.kt
jeffdgr8: https://github.com/Kotlin/kotlinx-io/pull/174#discussion_r1262841741

core/apple/src/SinksApple.kt
jeffdgr8: I named this `writeNative()` because it does what this equivalent [`Buffer.write()` function](https://github.com/Kotlin/kotlinx-io/blob/03845eaa0f62c4d54fc5f3ea0559e0176c2c5fa5/core/common/src/Buffer.kt#L395) does, but takes a native memory pointer for its `source` parameter, rather than Kotlin's `ByteArray`. Maybe it should just be called `write()`, since it's essentially overloading this function's behavior?

core/apple/src/SinksApple.kt
jeffdgr8: Yes, good catch.

core/apple/src/SinksApple.kt
whyoleg: Hm, you are right! I agree
Also, I would think, that in future this 'Buffer.write(pointer)' should become public, so we can already declare it top level, but internal, but it's not really necessary, just thoughts

core/apple/src/SinksApple.kt
jeffdgr8: Good catch! Yes, comparing to `NSOutputStream.outputStreamToMemory`'s behavior, it should not consume the buffer data, just return its current state as read-only.

jeffdgr8: Now the question is, is there an existing API to get the full `Sink`'s buffer state as a `ByteArray` without consuming it?

jeffdgr8: Ideally this should avoid copying the buffer multiple times as well, so probably need to implement this to natively copy directly to `NSData` anyway.

core/apple/src/SinksApple.kt
jeffdgr8: That's true. This would probably fit well in a `BuffersApple` file, similar to [`BuffersJvm`](https://github.com/Kotlin/kotlinx-io/blob/03845eaa0f62c4d54fc5f3ea0559e0176c2c5fa5/core/jvm/src/BuffersJvm.kt). I'll create the file and make it `internal` for now.

core/apple/src/SinksApple.kt
whyoleg: there is `Buffer.copyTo(Buffer)`, so I think similar could be implemented for `NSData`
But for now, it's possible to do `buffer.snapshot()` (will return ByteString), and then via unsafe access get underlying `ByteArray` and create `NSData` from it

core/apple/src/SourcesApple.kt
jeffdgr8: For `Source` reads, the [JVM implementation](https://github.com/Kotlin/kotlinx-io/blob/03845eaa0f62c4d54fc5f3ea0559e0176c2c5fa5/core/common/src/Buffer.kt#L326-L344) doesn't iterate over segments. [This test](https://github.com/Kotlin/kotlinx-io/pull/174/files#diff-8af6b3dce7132e2e2dba9eb3e0a373944f872248e24c670243a0e6bdbf5a665dR40) specifically tests for this behavior, which I initially borrowed from [this Okio test](https://github.com/square/okio/blob/bec51e206c7b95a01a9ccbe9eff126d28f2fac2b/okio/src/jvmTest/kotlin/okio/OkioTest.kt#L108). I don't see this same test in kotlinx-io though.

core/apple/test/NSOutputStreamSinkTest.kt
jeffdgr8: Interesting, I haven't seen this locally. It also passed on iOS, watchOS, and tvOS. Looks like a corrupted pointer coming from the `NSOutputStream`'s `NSStreamDataWrittenToMemoryStreamKey` `NSData` property. Now the question is why would that memory have been freed or was the data actually empty or was it corrupted sometime before? [I added](https://github.com/Kotlin/kotlinx-io/pull/174/files#diff-0dac6639d2e10ace0d178ca7cdfb08b54e570eba7dbf81cabf4603db62349cd4R22-R24) an additional check for the `NSData` length being 1 and changed the code accessing the byte from the byte array.

core/apple/src/AppleCore.kt
whyoleg: > Interestingly, you can't close an NSOutputStream that's never been opened. So you have to write something before you can close the Sink too.

So what behaviour we will have if we do `NSOutputStream.asSink().close()`? (so when stream is not opened)

core/apple/src/SinksApple.kt
whyoleg: I think that's the last point which is not done yet, right?

core/apple/src/SourcesApple.kt
whyoleg: for 2 I also don't have strong opinion, so up to @fzhinkin do decide I think :)

whyoleg (commented):
LGTM after a fix of `propertyForKey(NSStreamDataWrittenToMemoryStreamKey)` behaviour
Though, I've not reviewed tests, mostly just scrolled...

core/apple/test/NSOutputStreamSinkTest.kt
fzhinkin: I guess the issue caused by the reinterpret cast from int8* to int32*: the read following it fetches the expected single byte and three bytes after it.
Note that `-2145945247` is `0x80184961`, and given little-endian ordering on X86-64, the test reads:
```
0x61 0x49 0x18 0x80
  ^   ^
  |   |
  |   Garbage after it
  |
The value you're trying to verify
```

jeffdgr8: Ah, yes! Here I am writing a byte and reading four of them. Fixed that.

core/apple/src/AppleCore.kt
jeffdgr8: > So what behaviour we will have if we do NSOutputStream.asSink().close()? (so when stream is not opened)

The stream will remain in the `NSStreamStatusNotOpen` state. If we want to avoid this, we could `open()` and then `close()` the stream if it is in this state. That could end up doing extra work unnecessarily though, depending on the `NSOutputStream` implementation. Not sure which is preferable.

core/apple/src/SinksApple.kt
jeffdgr8: I implemented `Buffer.snapshotAsNSData()` to copy the buffer state to `NSData` directly. Yes, this should have been the last feedback item.

core/apple/src/BuffersApple.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.cinterop.*
+import kotlinx.io.bytestring.ByteString
+import kotlinx.io.bytestring.buildByteString
+import platform.Foundation.*
+import platform.darwin.NSUInteger
+import platform.darwin.NSUIntegerMax
+import platform.posix.memcpy
+import platform.posix.uint8_tVar
+
+internal fun Buffer.write(source: CPointer<uint8_tVar>?, maxLength: Int) {
```
fzhinkin, line 14: Please, verify that `maxLength` is non-negative and that `source` is not null.

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,108 @@
...
+    private var status = NSStreamStatusNotOpen
+    private var error: NSError? = null
+        set(value) {
+            status = NSStreamStatusError
+            field = value
+        }
+
+    private var pinnedBuffer: Pinned<ByteArray>? = null
+
+    override fun streamStatus() = if (isClosed()) NSStreamStatusClosed else status
```
fzhinkin, line 46: Is the `isClosed()` test here aimed for the cases when someone holding a reference to the underlying `source` will close it without calling `close` on the stream?
Should the same check be placed in the output stream's `streamStatus`?

core/apple/test/NSInputStreamSourceTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    @Test
+    fun sourceFromInputStreamBounds() {
+        val source = NSInputStream(ByteArray(100).toNSData()).asSource()
+        try {
+            source.readAtMostTo(Buffer(), -1)
+            fail()
+        } catch (expected: IllegalArgumentException) {
+            // expected
+        }
```
fzhinkin, line 69:
```suggestion
        assertFailsWith<IllegalArgumentException> { source.readAtMostTo(Buffer(), -1) }

```

core/apple/src/BuffersApple.kt
```diff
@@ -0,0 +1,67 @@
+@file:OptIn(UnsafeNumber::class)
+
+package kotlinx.io
+
+import kotlinx.cinterop.*
+import kotlinx.io.bytestring.ByteString
+import kotlinx.io.bytestring.buildByteString
```
fzhinkin, line 7:
```suggestion
```

core/apple/src/AppleCore.kt
fzhinkin: I see the issue with how closing is handled now; consider the following case:
```
val source = NSInputStream.inputStreamWithFileAtPath("/dev/random")!!.asSource()
source.close()
assertFailsWith<Throwable> {
        source.readAtMostTo(Buffer(), 10)
}
```
The test will fail because `close` is effectively doing nothing here (the stream is not opened yet) and the next read attempt will open the stream and it won't be considered as closed.

I'm fine with opening the stream on the first read as long as the behavior will be reflected in the class' documentation. But the close should be handled properly - if we closed the source before reading for it then the next read should fail.
The same is true for sinks.

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,108 @@
...
+    override fun streamError() = error
+
+    override fun open() {
+        if (status == NSStreamStatusNotOpen) {
+            status = NSStreamStatusOpen
+        }
+    }
+
+    override fun close() {
+        status = NSStreamStatusClosed
```
fzhinkin, line 57: At least for `NSInputStream.inputStreamWithFileAtPath` (well, it was the type of stream for which it was easy to check the case), the stream status remains `NSStreamStatusError` after the error happened and `close()` doesn't change it.
I'm not sure if it's explicitly described somewhere in Apple's documentation, but I would suggest mimicking the behavior of standard streams here.

core/apple/src/AppleCore.kt
```diff
@@ -0,0 +1,111 @@
...
+ */
+public fun NSOutputStream.asSink(): RawSink = OutputStreamSink(this)
+
+private open class OutputStreamSink(
+    private val out: NSOutputStream,
+) : RawSink {
+
+    @OptIn(UnsafeNumber::class)
+    override fun write(source: Buffer, byteCount: Long) {
+        if (out.streamStatus == NSStreamStatusClosed) throw IOException("Stream Closed")
```
fzhinkin, line 30: Should we check other possible status values here?
For example, check if it's `NSStreamStatusWriting` and throw an exception in that case.

core/apple/src/BuffersApple.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    return toCopy
+}
+
+internal fun Buffer.snapshotAsNSData(): NSData {
+    if (size == 0L) return NSData.data()
+
+    check(size.toULong() <= NSUIntegerMax) { "Buffer is too long ($size) to be converted into NSData." }
+
+    val data = NSMutableData.create(length = size.convert())!!
```
fzhinkin, line 53: Does it make sense to allocate a block of memory, fill it and then return wrap into `NSData` (https://developer.apple.com/documentation/foundation/nsdata/1416020-init) instead of creating and returning `NSMutableData`?
Or can one always cast a pointer to `NSData` to `NSMutableData` and modify the underlying data anyway? I'm sorry for my ignorance about that.

core/apple/test/SourceNSInputStreamTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import platform.darwin.NSUIntegerVar
+import platform.darwin.UInt8Var
+import kotlin.test.*
+
+@OptIn(UnsafeNumber::class)
+class SourceNSInputStreamTest {
+    @Test
+    fun bufferInputStream() {
+        val source = Buffer()
+        source.writeString("abc")
```
fzhinkin, line 22: Please add some cases with longer data written into the buffer to cover the case when multiple segments are involved in reading from a source.

fzhinkin (changes requested):

core/build.gradle.kts
```diff
@@ -53,8 +53,10 @@ kotlin {
         val jvmMain by getting
         val jvmTest by getting
 
-        createSourceSet("nativeMain", parent = commonMain, children = nativeTargets)
-        createSourceSet("nativeTest", parent = commonTest, children = nativeTargets)
+        val nativeMain = createSourceSet("nativeMain", parent = commonMain, children = nativeTargets)
+        val nativeTest = createSourceSet("nativeTest", parent = commonTest, children = nativeTargets)
+        createSourceSet("appleMain", parent = nativeMain, children = appleTargets)
+        createSourceSet("appleTest", parent = nativeTest, children = appleTargets)
```
fzhinkin, line 59: BTW, the same change is needed in the bytestring's `build.gradle.kts`. There are no native sources to cause issues right now, but it would be nice to prevent them from appearing in the future.

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,85 @@
...
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENCE file.
+ */
+
+package kotlinx.io
+
+import kotlinx.cinterop.*
+import platform.Foundation.*
+import platform.darwin.NSInteger
+import platform.darwin.NSUInteger
+import platform.posix.memcpy
```
fzhinkin, line 12:
```suggestion
```

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,85 @@
...
+    private val isClosed: () -> Boolean = when (sink) {
+        is RealSink -> sink::closed
+        is Buffer -> {
+            { false }
+        }
+    }
+
+    @OptIn(ExperimentalStdlibApi::class)
+    @Volatile
+    private var status = NSStreamStatusNotOpen
```
fzhinkin, line 37: Why is the status volatile?
It seems like NSStream doesn't provide any thread-safety guarantees.

core/apple/test/NSInputStreamSourceTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import platform.Foundation.NSInputStream
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.fail
+
+class NSInputStreamSourceTest {
+    @Test
+    fun nsInputStreamSource() {
+        val input = NSInputStream(byteArrayOf(0x61).toNSData())
```
fzhinkin, line 16: Different kinds of NSInputStreams (files, in-memory, etc) have different implementations, its worth checking that we correctly handle all of them, not only array-backed on.

I guess it may not be that easy to check socket-based one, but for files it should be relatively simple.

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,108 @@
...
+ * Returns an input stream that reads from this source. Closing the stream will also close this source.
+ *
+ * @sample kotlinx.io.samples.KotlinxIoSamplesApple.asStream
+ */
+public fun Source.asNSInputStream(): NSInputStream = SourceNSInputStream(this)
+
+@OptIn(InternalIoApi::class, UnsafeNumber::class)
+private class SourceNSInputStream(
+    private val source: Source,
+) : NSInputStream(NSData()) {
```
fzhinkin, line 26: `SourceNSInputStream` does not implement `scheduleInRunLoop`, so there will an error on attempt to execute code like this:
```
val stream = Buffer().also { it.writeString("blablablabla") }.asNSInputStream()
        stream.scheduleInRunLoop(NSRunLoop.currentRunLoop(), NSRunLoopCommonModes)
```

```
 2023-07-17 19:00:27.543 test.kexe[60305:7494596] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -scheduleInRunLoop:forMode: only defined for abstract class.  Define -[Test_kobjc0 scheduleInRunLoop:forMode:]!'
```

And some modern frameworks (like `NSURLSession`) still work with run loops. 

I'm not an ObjC/Swift expert, but given what I was told, inheriting form `NSInputStream`/`NSOutputStream` seems tricky.

core/apple/src/AppleCore.kt
jeffdgr8: So should we `open()` and then `close()` the underlying `NSInputStream` or `NSOutputStream` if closed before read or write? Or should we `open()` it when the `Source` or `Sink` is created in `init`?

fzhinkin: Personally, I like the idea of opening the stream in the `init` more.

jeffdgr8: I think that makes more sense to me as well, now that we've iterated through the behavior of delaying the opening until first read/write, since essentially we're saying the stream must be opened eventually however you end up using it (assuming valid usage should always close the stream). The only benefit to delaying opening might be to prevent potential resource leakage if a user creates a stream but never reads/writes or closes the stream, which should be unexpected. I'll go ahead and revert to opening on `init`.

core/apple/test/utilApple.kt
```diff
@@ -0,0 +1,12 @@
+package kotlinx.io
```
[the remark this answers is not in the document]
jeffdgr8: Did you mean the `samplesApple.kt` file? Like [this](https://github.com/Kotlin/kotlinx-io/pull/174/commits/4e3ff8786afe1a42c8b88c32001ed3e6fc591cfc#diff-a6fced4b4e4a99784a2b92028b1adbc8c0ca28ff54e264b0e73727b3d4a460cfR93)?

core/apple/src/BuffersApple.kt
jeffdgr8: I made `source` non-optional here and did the check in the caller and also applied these changes to `Buffer.readAtMost()`.

core/apple/src/SourcesApple.kt
jeffdgr8: Yes, that was my reasoning. I'll add to `SinkNSOutputStream` as well.

core/apple/src/SourcesApple.kt
jeffdgr8: That's a great point. I've noticed there are a number of subtle behaviors that aren't explicitly documented by Apple. It seems the terminal statuses for an `NSStream`, where a stream can no longer be used or updated to another status, are either `NSStreamStatusClosed` or `NSStreamStatusError`. Any temporary or parameter-related error doesn't update the `streamError` property or status.

Along these same lines, we should probably close the underlying `Source` when the `NSInputStream` is terminated do to an error. I fixed this for both `SourceNSInputStream` and `SinkNSOutputStream`.

core/apple/src/AppleCore.kt
jeffdgr8: Other status errors should cause an exception to be thrown on [this line](https://github.com/Kotlin/kotlinx-io/pull/174/files#diff-5bb5a6a76d27b1ed60b0526e03021a31b1b429883dd0017726c83edacf44347aR47). This handles `NSStreamStatusError` with an error message from the underlying `NSError`. However, a concurrency error would likely result in `Unknown error`, since it's not a terminal error.

Since we're not providing thread safety guarantees, checking for `NSStreamStatusWriting` here might give the illusion of thread safety. A successful check at the beginning of the function could still result in a failed `write()` call, if the other thread hasn't begun the `write()` operation until right after the check. Similarly a failed check would have no guarantee the `write()` call wouldn't have succeeded, had the other thread completed before we actually got to the call. And in either case, it's likely there'd be buffer state corruption by updating its state concurrently.

core/apple/src/BuffersApple.kt
jeffdgr8: You can't necessarily cast any `NSData` as `NSMutableData` to modify it. You could get a cast exception. So yes, initializing `NSData` with pre-allocated bytes would prevent the cast. Although since this is a snapshot copy of the buffer state, there's not really an issue if the caller did decide to modify the bytes, since it's a copy they own. It's also always possible to modify the underlying bytes of `NSData` with lower-level APIs, even if it is supposed to be immutable.

But another advantage of allocating the bytes with `malloc` would be that it doesn't bother zeroing them out as [`NSMutableData.dataWithLength:`](https://developer.apple.com/documentation/foundation/nsmutabledata/1547233-datawithlength#return_value) does, so this could have a slight performance advantage. I'll go ahead and make that change.

core/apple/test/SourceNSInputStreamTest.kt
jeffdgr8: I added these longer input data test cases: https://github.com/Kotlin/kotlinx-io/pull/174/commits/365c35421ce4402d5127abc3fca6478edb5b54cd

core/apple/src/SinksApple.kt
jeffdgr8: Yes, despite not providing thread safety guarantees for operating on the streams, I just didn't see much use for the status being updated to `NSStreamStatusReading` and `NSStreamStatusWriting` if this wasn't at least visible to another thread from a read-only perspective. The thread doing the reading and writing would never see the status change, since it's in the middle of the read and write functions.

I suppose a caller could use key-value coding to observe the `streamStatus` property and the observer method would be called on the same thread that updated the status in the middle of the read or write. This might actually be the only really valid use case for those statuses, since it's likely not feasible to poll the status from another thread.

Are there reasons to not make the property volatile? If so, should I remove the `@Volatile` annotation then?

core/apple/test/NSInputStreamSourceTest.kt
jeffdgr8: I added this `NSInputStream` from file URL test: https://github.com/Kotlin/kotlinx-io/pull/174/commits/002141238f5a3ac19f309bb47436160f6912cea9

I'm confused why this works though, as the `file` path string is always 1 or 2 bytes of seemingly garbage (not characters), but somehow it creates a `NSInputStream` from this file `NSURL` and reads what the `Sink` wrote to the file. 🤔

jeffdgr8: I guess it is just creating a file with the junk character(s) name in the current working directory. Probably need to revisit `createTempFile()` for Apple platforms. `mktemp()` doesn't always return null, but it doesn't seem to ever return something valid.

Also, `"$tmpDir/tmp-${rnd.nextInt()}"` puts a double `//` in the path: `...//tmp-...`.

core/apple/src/SourcesApple.kt
jeffdgr8: I did note this, but I haven't looked into implementing this `NSStream` API, as my use-case and others that just use `NSInputStream`'s `read()` API directly, don't require it. I can look into implementing, if it's not too much work.

Can you give an example of how this API is used in `NSURLSession`?

core/apple/test/NSInputStreamSourceTest.kt
jeffdgr8: I created https://github.com/Kotlin/kotlinx-io/issues/183 to track this.

fzhinkin: Would you consider replacing `createTempFile` with `"${NSTemporaryDirectory()}${NSUUID().UUIDString()}"` for now?

I don't want fixing #183 concurrently with your PR to avoid additional merge conflicts in build scripts (as native test sources should be split into appleTest and nonAppleTest) and would rather fix it after.

core/apple/src/SinksApple.kt
jeffdgr8: That makes sense. I'll remove the `@Volatile` annotations.

core/apple/test/NSInputStreamSourceTest.kt
jeffdgr8: I made the change and left a comment about the future fix.

core/apple/src/SourcesApple.kt
jeffdgr8: Great, thanks for the test case. I'll look into implementing the delegate and run loop functionality. I ran across that blog post (not a lot written on the subject of `NSInputStream` subclasses), but didn't read it fully at the time. Interestingly, the author, BJ, is a local iOS dev I'm in a Slack group with. Maybe I'll reach out to him if I run into questions.

jeffdgr8: I implemented the `NSStreamDelegate` and `scheduleInRunLoop()` APIs for both `SourceNSInputStream` and `SinkNSOutputStream`. Let me know if there's anything else I've missed.

fzhinkin: Thank you for fixing it, that's a terrific job!

fzhinkin: But it will take some time for me to get into details and review the change.

fzhinkin:
There are a few test failures that happened in CI:
1. `kotlinx.io.SourceNSInputStreamTest.delegateTest[iosX64]` failed w/ timeout:
```
kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 5000 ms
  at kotlin.Throwable#<init>(/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Throwable.kt:25)
  at kotlin.Exception#<init>(/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:23)
  at kotlin.RuntimeException#<init>(/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:34)
  at kotlin.IllegalStateException#<init>(/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:70)
  at kotlin.coroutines.cancellation.CancellationException#<init>(/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/coroutines/cancellation/CancellationException.kt:11)
  at kotlinx.coroutines.TimeoutCancellationException#<init>(/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/Timeout.kt:172)
  at kotlinx.coroutines#TimeoutCancellationException(/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/Timeout.kt:191)
  at kotlinx.coroutines.TimeoutCoroutine.run#internal(/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/Timeout.kt:159)
  at kotlinx.coroutines.WorkerDispatcher.DisposableBlock.invoke#internal(/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt:49)
  at kotlinx.coroutines.WorkerDispatcher.DisposableBlock.$<bridge-UNN>invoke(/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt:48)
  at <global>.WorkerLaunchpad(/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/native/concurrent/Internal.kt:87)
  at <global>._ZN6Worker19processQueueElementEb(Unknown Source)
  at <global>._ZN12_GLOBAL__N_113workerRoutineEPv(Unknown Source)
  at <global>._pthread_start(Unknown Source)
  at <global>.thread_start(Unknown Source)
  ```
  
 2. `kotlinx.io.SourceNSInputStreamTest.testRunLoopSwitch[iosX64]` failed w/:
 ```
------- Stdout: -------
Uncaught Kotlin exception: kotlin.AssertionError: Expected <run-loop-1>, actual <run-loop-2>.
Invalid connection: com.apple.coresymbolicationd
    at 0   test.kexe                           0x1018ce00b        kfun:kotlin.Throwable#<init>(kotlin.String?){} + 107 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Throwable.kt:25:37)
    at 1   test.kexe                           0x1018c7b87        kfun:kotlin.Error#<init>(kotlin.String?){} + 103 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:12:44)
Child process terminated with signal 6: Abort trap
    at 2   test.kexe                           0x1018c8c3d        kfun:kotlin.AssertionError#<init>(kotlin.Any?){} + 285 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:128:41)
    at 3   test.kexe                           0x10192a1f3        kfun:kotlin.test.DefaultAsserter#fail(kotlin.String?){}kotlin.Nothing + 275 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/DefaultAsserter.kt:16:19)
    at 4   test.kexe                           0x101926d41        kfun:kotlin.test.Asserter#assertTrue(kotlin.Function0<kotlin.String?>;kotlin.Boolean){} + 369 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt:652:13)
    at 5   test.kexe                           0x101927251        kfun:kotlin.test.Asserter#assertEquals(kotlin.String?;kotlin.Any?;kotlin.Any?){} + 657 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt:671:9)
    at 6   test.kexe                           0x101925ec7        kfun:kotlin.test#assertEquals(0:0;0:0;kotlin.String?){0§<kotlin.Any?>} + 391 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt:63:14)
    at 7   test.kexe                           0x101926094        kfun:kotlin.test#assertEquals$default(0:0;0:0;kotlin.String?;kotlin.Int){0§<kotlin.Any?>} + 356 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt:62:1)
    at 8   test.kexe                           0x1016f0cce        kfun:kotlinx.io.SourceNSInputStreamTest.object-3.stream#internal + 1166 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/test/SourceNSInputStreamTest.kt:273:25)
    at 9   test.kexe                           0x1016f4526        _6f72672e6a6574627261696e732e6b6f746c696e783a6b6f746c696e782d696f2d636f72655f746573742f6f70742f6275696c644167656e742f776f726b2f323265336137646236623836316261372f636f72652f6170706c652f746573742f536f757263654e53496e70757453747265616d546573742e6b74_knbridge107 + 246 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/test/SourceNSInputStreamTest.kt:271:26)
    at 10  test.kexe                           0x1016ac0f6        kfun:kotlinx.io.SourceNSInputStream.checkBytes$lambda$3#internal + 2806 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/src/SourcesApple.kt:124:31)
    at 11  test.kexe                           0x1016ace02        kfun:kotlinx.io.SourceNSInputStream.$checkBytes$lambda$3$FUNCTION_REFERENCE$4.invoke#internal + 82 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/src/SourcesApple.kt:111:17)
    at 12  test.kexe                           0x1016acf4f        kfun:kotlinx.io.SourceNSInputStream.$checkBytes$lambda$3$FUNCTION_REFERENCE$4.$<bridge-UNN>invoke(){}#internal + 63 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/src/SourcesApple.kt:111:17)
    at 13  test.kexe                           0x1016ad331        _6f72672e6a6574627261696e732e6b6f746c696e783a6b6f746c696e782d696f2d636f72652f6f70742f6275696c644167656e742f776f726b2f323265336137646236623836316261372f636f72652f6170706c652f7372632f536f75726365734170706c652e6b74_knbridge5 + 209 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/src/SourcesApple.kt:111:17)
    at 14  CoreFoundation                      0x7fff20369bba     __CFRUNLOOP_IS_CALLING_OUT_TO_A_BLOCK__ + 11
    at 15  CoreFoundation                      0x7fff20368fc0     __CFRunLoopDoBlocks + 442
    at 16  CoreFoundation                      0x7fff20363e14     __CFRunLoopRun + 2094
    at 17  CoreFoundation                      0x7fff20363102     CFRunLoopRunSpecific + 566
    at 18  Foundation                          0x7fff2081941b     -[NSRunLoop(NSRunLoop) runMode:beforeDate:] + 212
    at 19  Foundation                          0x7fff20819637     -[NSRunLoop(NSRunLoop) run] + 75
    at 20  test.kexe                           0x1016f8108        kfun:kotlinx.io.startRunLoop$lambda$0#internal + 1160 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/test/utilApple.kt:39:17)
    at 21  test.kexe                           0x1016f9366        kfun:kotlinx.io.$startRunLoop$lambda$0$FUNCTION_REFERENCE$0.invoke#internal + 70 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/test/utilApple.kt:35:5)
    at 22  test.kexe                           0x1016f947f        kfun:kotlinx.io.$startRunLoop$lambda$0$FUNCTION_REFERENCE$0.$<bridge-UNN>invoke(){}#internal + 63 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/test/utilApple.kt:35:5)
    at 23  test.kexe                           0x1016f9581        _6f72672e6a6574627261696e732e6b6f746c696e783a6b6f746c696e782d696f2d636f72655f746573742f6f70742f6275696c644167656e742f776f726b2f323265336137646236623836316261372f636f72652f6170706c652f746573742f7574696c4170706c652e6b74_knbridge6 + 209 (/opt/buildAgent/work/22e3a7db6b861ba7/core/apple/test/utilApple.kt:35:5)
    at 24  Foundation                          0x7fff20842141     __NSThread__start__ + 998
    at 25  libsystem_pthread.dylib             0x7fff6bfee8fb     _pthread_start + 223
    at 26  libsystem_pthread.dylib             0x7fff6bfea442     thread_start + 14
 ```

**Comment by jeffdgr8** [comment]
> There are a few test failures that happened in CI:

Thanks, I'll take a look. Glad the CI seems to create better failure conditions than my local machine.

The only test I noted would occasionally hang locally was the upload test. I've noticed this with some Kotlin/Native tests if a background thread is still active when the main test function completes. So might be related to this.

I pushed up some changes that will log the timeout failure more precisely. Hopefully the CI can reproduce it again.

**Comment by jeffdgr8** [comment]
I pushed a fix that should address the `Expected <run-loop-1>, actual <run-loop-2>` failure. It was a flaw in the test, where the newly scheduled run loop could schedule its `NSStreamEventHasBytesAvailable` event before the `read` variable had been updated.

File: core/apple/src/SourcesApple.kt [hunk 28]
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private fun checkBytes(sendEndEvent: Boolean = true) {
+        val runLoop = runLoop ?: return
+        runLoop.performInModes(runLoopModes) {
+            if (runLoop != this.runLoop) return@performInModes
+            try {
+                if (source.exhausted()) {
+                    if (sendEndEvent) delegate?.stream(this, NSStreamEventEndEncountered)
+                    val timer = NSTimer.timerWithTimeInterval(0.1, false) {
```
jeffdgr8, line 116: Perhaps it'd make sense to use a back off algorithm here instead of a fixed poll interval, since the frequency of byte availability from a given `Source` is unknown and would vary with different use cases.

Alternatively, the poll interval could be made configurable, possibly with a key value passed to `setProperty()`.

core/apple/test/SourceNSInputStreamTest.kt
```diff
@@ -0,0 +1,354 @@
...
+            }
+            input.open()
+            input.scheduleInRunLoop(runLoop, NSDefaultRunLoopMode)
+            runBlocking {
+                available.lockWithTimeout()
+            }
+            input.close()
+        }
+
+        subscribeAfterOpen(NSInputStream(data = "default".encodeToByteArray().toNSData()))
```
fzhinkin, line 350: It's reasonable to compare the behavior of custom and default stream implementation during development and debugging, but I don't think we need to run tests for default Foundation streams on a regular basis.

core/apple/test/SourceNSInputStreamTest.kt
```diff
@@ -0,0 +1,354 @@
...
+            println("subscribeAfterOpen(${input::class})")
+            val available = Mutex(true)
+
+            input.delegate = object : NSObject(), NSStreamDelegateProtocol {
+                override fun stream(aStream: NSStream, handleEvent: NSStreamEvent) {
+                    println("${input::class} event = ${handleEvent.asString()} (${NSThread.currentThread.name})")
+                    assertEquals("run-loop", NSThread.currentThread.name)
+                    when (handleEvent) {
+                        NSStreamEventOpenCompleted -> fail("opened before subscribe")
+                        NSStreamEventHasBytesAvailable -> available.unlock()
```
fzhinkin, line 338: Does it make sense to check what data is there?

core/apple/test/SourceNSInputStreamTest.kt
```diff
@@ -0,0 +1,354 @@
...
+
+        fun subscribeAfterOpen(input: NSInputStream) {
+            println("subscribeAfterOpen(${input::class})")
+            val available = Mutex(true)
+
+            input.delegate = object : NSObject(), NSStreamDelegateProtocol {
+                override fun stream(aStream: NSStream, handleEvent: NSStreamEvent) {
+                    println("${input::class} event = ${handleEvent.asString()} (${NSThread.currentThread.name})")
+                    assertEquals("run-loop", NSThread.currentThread.name)
+                    when (handleEvent) {
```
fzhinkin, line 336: Should the test fail on other events?

core/apple/test/SourceNSInputStreamTest.kt
```diff
@@ -0,0 +1,354 @@
...
+            println("consumeSwitching(${input::class}, $data)")
+            val opened = Mutex(true)
+            val readLock = reentrantLock()
+            var read = 0
+            val completed = Mutex(true)
+
+            input.delegate = object : NSObject(), NSStreamDelegateProtocol {
+                val sink = ByteArray(data.length)
+                override fun stream(aStream: NSStream, handleEvent: NSStreamEvent) {
+                    readLock.withLock {
```
fzhinkin, line 274: These tests use a relatively complex locking scheme to ensure the ordering and correctness of different phases of the stream's processing. It would be nice to have a comment describing the expected phase transition and locking order.

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,137 @@
...
+ * Returns an output stream that writes to this sink. Closing the stream will also close this sink.
+ *
+ * @sample kotlinx.io.samples.KotlinxIoSamplesApple.asStream
+ */
+public fun Sink.asNSOutputStream(): NSOutputStream = SinkNSOutputStream(this)
+
+@OptIn(UnsafeNumber::class)
+private class SinkNSOutputStream(
+    private val sink: Sink
+) : NSOutputStream(toMemory = Unit), NSStreamDelegateProtocol {
```
fzhinkin, line 25: I'm not sure how things work under the hood when inheriting NSInputStream/NSOutputStream and calling a non-default initializer (I understand that no-arg initializer is not available here) as according to Apple's docs, these initializers aimed to instantiate some specific stream's subclass.

fzhinkin (changes requested):

core/apple/src/SourcesApple.kt
jeffdgr8: I was thinking there may potentially be other `Source` implementations that could have bytes become available in the future, like a socket protocol. I'd assume the implementation would ensure thread safety either internally or by scheduling itself on the same run loop as the consumer. This is correct though, that my test using `Buffer` is not thread safe. It does make sense `NSStreamEventEndEncountered` could be considered a terminal event to close the stream. I can just remove the polling.

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,164 @@
...
+                    }
+                    runLoopModes.forEach { mode ->
+                        runLoop.addTimer(timer, mode)
+                    }
+                } else {
+                    status = NSStreamStatusOpen
+                    delegate?.stream(this, NSStreamEventHasBytesAvailable)
+                }
+            } catch (e: IllegalStateException) {
+                // ignore closed
```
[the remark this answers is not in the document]
jeffdgr8: My intention was to avoid crashing with a possible race condition if the stream was closed on another thread immediately after the run loop thread checked the closed state. I would think typical usage when using the run loop would be to close the stream on the run loop thread after receiving the `NSStreamEventEndEncountered` event, but there might be a reason a consumer would prematurely close the stream and wouldn't want to crash here if that were the case.

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,137 @@
...
+    @OptIn(DelicateIoApi::class)
+    override fun write(buffer: CPointer<uint8_tVar>?, maxLength: NSUInteger): NSInteger {
+        try {
+            if (isClosed()) throw IOException("Underlying sink is closed.")
+            if (status != NSStreamStatusOpen) return -1
+            if (buffer == null) return -1
+
+            status = NSStreamStatusWriting
+            sink.writeToInternalBuffer {
+                it.write(buffer, maxLength.toInt())
```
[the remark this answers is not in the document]
jeffdgr8: Since the internal buffer segment size is measured with `Int`, truncating to `Int.MAX_VALUE` in this case makes sense.

core/apple/test/SourceNSInputStreamTest.kt
```diff
@@ -0,0 +1,354 @@
...
+            assertEquals("[-5, -5, -5, -5]", byteArray.contentToString())
+        }
+    }
+
+    @Test
+    fun delegateTest() {
+        val runLoop = startRunLoop()
+
+        fun consumeWithDelegate(input: NSInputStream, data: String) {
+            println("consumeWithDelegate(${input::class}, $data)")
```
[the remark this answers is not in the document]
jeffdgr8: This was for debugging purposes with the test that was failing in the CI. I'll need to find a better way to track the specific function call that failed if it's not logged, since coroutine stack traces are truncated, you can't see the original caller to know which stream caused the failure.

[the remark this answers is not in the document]
jeffdgr8: After removing the platform stream tests, most of the functions are only called once. So hopefully there isn't a need for these further debug details.

core/apple/test/SourceNSInputStreamTest.kt
jeffdgr8: My intention for this test was just checking that the `NSStreamEventOpenCompleted` event isn't sent when the stream is opened before scheduling in the run loop. So just getting the `NSStreamEventHasBytesAvailable` event instead was all that was necessary to test this case. But I can add a read check as well, for good measure.

core/apple/test/SourceNSInputStreamTest.kt
jeffdgr8: Now after adding the data check after read, it will get the `NSStreamEventEndEncountered` event, which as long as it's in the correct order is expected. I'll add explicit checks for the other events to be thorough.

core/apple/src/SinksApple.kt
jeffdgr8: This was the only way I found to appease the requirement `Unable to call non-designated initializer as super constructor`, which is the error the compiler throws if you use the no-arg constructor. This seems to be a conflict between the way Kotlin and Objective-C object inheritance works, where Objective-C can return a subclass implementation as `self` from an `init` function.

I'm not sure there really should be an expected designated initializer in Kotlin for either of these classes. This is likely an interop error. In Objective-C this compiles and I'm able to construct `[[MyNSInputStream alloc] init]` just fine:

```
// MyNSInputStream.h
@interface MyNSInputStream : NSInputStream
- (instancetype)init;
@end

// MyNSInputStream.m
@implementation MyNSInputStream
- (instancetype)init {
    self = [super init];
    return self;
}
@end
```

It works to use either the memory or URL constructor of both `NSInputStream` and `NSOutputStream`. I figured the empty memory versions were likely to be the most lightweight. Is there some way to suppress this designated initializer requirement to call the no-arg constructor? It might also make sense to file an issue to fix the interop. Since we're overriding all of `NSInputStream` and `NSOutputStream`'s public API functions and not using anything from the superclass, I think this should be safe though.

core/apple/test/SourceNSInputStreamTest.kt
```diff
@@ -0,0 +1,354 @@
...
+
+        consumeWithDelegate(NSInputStream(data = "default".encodeToByteArray().toNSData()), "default")
+        consumeWithDelegate(Buffer().apply { writeString("custom") }.asNSInputStream(), "custom")
+        consumeWithDelegate(NSInputStream(data = NSData.data()), "")
+        consumeWithDelegate(Buffer().asNSInputStream(), "")
+        CFRunLoopStop(runLoop.getCFRunLoop())
+    }
+
+    @Test
+    fun uploadTest() {
```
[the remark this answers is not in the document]
jeffdgr8: It was a good test for verifying a real world use case, but probably not necessary since we've covered the delegate lifecycle in other tests. I'll just remove it.

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,164 @@
...
+            return read
+        } catch (e: Exception) {
+            error = e.toNSError()
+            return -1
+        }
+    }
+
+    override fun getBuffer(buffer: CPointer<CPointerVar<uint8_tVar>>?, length: CPointer<NSUIntegerVar>?) = false
+
+    override fun hasBytesAvailable() = !source.exhausted()
```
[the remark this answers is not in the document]
jeffdgr8: Good call. I changed it to `streamStatus != NSStreamStatusClosed && status != NSStreamStatusError`, since some sources can't be closed (`Buffer`), but we'll still close the `NSInputStream`.

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,137 @@
...
+    private fun postEvent(event: NSStreamEvent) {
+        val runLoop = runLoop ?: return
+        runLoop.performInModes(runLoopModes) {
+            if (runLoop == this.runLoop) {
+                delegate?.stream(this, event)
+            }
+        }
+    }
+
+    override fun delegate() = _delegate.value
```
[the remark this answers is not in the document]
jeffdgr8: Yeah, the docs are really not clear on this. And despite its [explicit requirement](https://developer.apple.com/documentation/foundation/nsstream/1418423-delegate#discussion) to make the stream its own delegate by default, this doesn't actually make sense with our subclass since our self delegate implementation does nothing, and our class isn't `open` for subclassing, in fact it's not even public. But that's the only reason I've implemented `NSStreamDelegateProtocol`, since it says to, and maybe there's some messaging magic that may be possible while using the Kotlin object within Objective-C code.

I checked the default `NSOutputStream.delegate` returns `null` by default. So I don't think the expectation is to return `self` when it's `nil`.

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,137 @@
...
+) : NSOutputStream(toMemory = Unit), NSStreamDelegateProtocol {
+
+    private val isClosed: () -> Boolean = when (sink) {
+        is RealSink -> sink::closed
+        is Buffer -> {
+            { false }
+        }
+    }
+
+    private var status = NSStreamStatusNotOpen
```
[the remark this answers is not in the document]
jeffdgr8: If the stream is not open, `write()` will just return -1, but not change the status to error. `NSStreamStatusError` is a result of a terminal error, so any error that is correctable, like calling a function in the wrong order or with bad parameters, should not change the status to error.

I don't think we should go from `NSStreamStatusClosed` to `NSStreamStatusError`. Calling `write` on a closed platform stream just returns -1. I'll change that throwing `IOException` to returning -1 when closed.

We should probably mirror the behavior of the platform streams and not allow closing a stream that has never been opened. I'll make that change. You're missing the close from the `NSStreamStatusOpen` state though.

Where are you getting the `NSStreamStatusOpen -->|error| NSStreamStatusError;` from? I believe we should only get an `error` during `write`. That's the only place it's being set.

I guess we just pass right through `NSStreamStatusOpening`. Maybe it makes sense to set it right before setting to `NSStreamStatusOpen`, in case some key-value observer is expecting it in Objective-C? We don't actually do anything to open the underlying `Sink` or `Source`, so setting both statuses is entirely to align with `NSStream` expected behavior.

Sinks don't provide a mechanism to be limited in capacity, so we never get `NSStreamStatusAtEnd`.

This is what I believe the status lifecycle should look like after these changes:

```mermaid
graph TD
    NSStreamStatusNotOpen -->|open| NSStreamStatusOpening;
    NSStreamStatusOpening -->|open| NSStreamStatusOpen;
    NSStreamStatusOpen -->|close| NSStreamStatusClosed;
    NSStreamStatusOpen -->|write| NSStreamStatusWriting;
    NSStreamStatusWriting -->|write| NSStreamStatusOpen;

    NSStreamStatusWriting -->|error| NSStreamStatusError;

    NSStreamStatusReading;
    NSStreamStatusAtEnd;
```

core/apple/src/SinksApple.kt
```diff
@@ -0,0 +1,137 @@
...
+        status = NSStreamStatusClosed
+        sink.close()
+        runLoop = null
+        runLoopModes = listOf()
+    }
+
+    @OptIn(DelicateIoApi::class)
+    override fun write(buffer: CPointer<uint8_tVar>?, maxLength: NSUInteger): NSInteger {
+        try {
+            if (isClosed()) throw IOException("Underlying sink is closed.")
```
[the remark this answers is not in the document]
jeffdgr8: I guess we don't have any checked exceptions being thrown now. The goal of the try/catch is to ensure we don't throw any exception during the I/O operation, since the `NSStream` API expects errors to be propagated through the `streamError` property. I narrowed the try/catch to the `write` operation. I'd think the most likely place we could get an exception would probably be with the `hintEmit` call in `writeToInternalBuffer`, which may flush data to the underlying `Sink` and might cause an `IOException`.

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,164 @@
...
+) : NSInputStream(NSData()), NSStreamDelegateProtocol {
+
+    private val isClosed: () -> Boolean = when (source) {
+        is RealSource -> source::closed
+        is Buffer -> {
+            { false }
+        }
+    }
+
+    private var status = NSStreamStatusNotOpen
```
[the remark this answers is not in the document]
jeffdgr8: I believe this should be the status lifecycle after updates:

```mermaid
graph TD
    NSStreamStatusNotOpen -->|open| NSStreamStatusOpening;
    NSStreamStatusOpening -->|open| NSStreamStatusOpen;
    NSStreamStatusOpen -->|read| NSStreamStatusReading;
    NSStreamStatusReading -->|read| NSStreamStatusAtEnd;
    NSStreamStatusReading --> |read| NSStreamStatusOpen;
    NSStreamStatusReading -->|error| NSStreamStatusError;
    NSStreamStatusAtEnd -->|close| NSStreamStatusClosed;
    NSStreamStatusOpen -->|close| NSStreamStatusClosed;
    
    NSStreamStatusWriting;
```

core/apple/src/SourcesApple.kt
fzhinkin: > I was thinking there may potentially be other Source implementations that could have bytes become available in the future, like a socket protocol.

That's something one can actually anticipate, but the Source's documentation explicitly states that `exhausted` means that the source is actually exhausted and no data is expected once `Source::exhausted` returned true. Source's underlying source is considered exhausted when a read form it returned `-1` and that particular value should only be returned if there will be no more data in the future (and if we expect to receive more bytes then the read should return `0`).
In case with sockets, it's up to an implementor how it'll work internally, but corresponding source's `readAtMostTo` should return `-1` only when a connection was closed, I guess.
So it's better to expect that particular `RawSource`/`RawSink` implementations correctly implement and obey the contract, instead of trying to handle their misbehavior.
`Buffer` is the outlier here, I can say nothing to defend its behavior. :)

core/apple/src/SinksApple.kt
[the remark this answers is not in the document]
fzhinkin: > Where are you getting the NSStreamStatusOpen -->|error| NSStreamStatusError; from? I believe we should only get an error during write. That's the only place it's being set.

That's possible when the stream is still open, but the underlying sink is closed and someone is calling `write`.
Not saying that it's wrong or problematic, just mentioning the possible transition.

> I guess we just pass right through NSStreamStatusOpening. Maybe it makes sense to set it right before setting to NSStreamStatusOpen, in case some key-value observer is expecting it in Objective-C?

My understanding is that it's mainly intended for streams with complex opening protocols, like network connections.

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,165 @@
...
+    override fun close() {
+        if (status == NSStreamStatusError || status == NSStreamStatusNotOpen) return
+        status = NSStreamStatusClosed
+        runLoop = null
+        runLoopModes = listOf()
+        source.close()
+    }
+
+    override fun read(buffer: CPointer<uint8_tVar>?, maxLength: NSUInteger): NSInteger {
+        if (streamStatus != NSStreamStatusOpen && streamStatus != NSStreamStatusAtEnd || buffer == null) return -1
```
fzhinkin, line 66: Sorry, I didn't notice it earlier: if we're in the `NSStreamStatusAtEnd` state, does it make sense to exit and abstain for calling `exhausted`?
The only way to transition into the at-end state is by exhausting the source and once it exhausted it's always exhausted (module Buffer's behavior).

core/apple/src/SinksApple.kt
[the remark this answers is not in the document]
jeffdgr8: > That's possible when the stream is still open, but the underlying sink is closed and someone is calling write.
Not saying that it's wrong or problematic, just mentioning the possible transition.

Oh, ok. This should be changed now where any `error` would come after `write` has changed status to `NSStreamStatusWriting`.

core/apple/src/SourcesApple.kt
jeffdgr8: Handling the behavior of `Buffer` or potentially another `Source` that behaved similarly is why I treated `NSStreamStatusAtEnd` as not necessarily a terminal state, like `NSStreamStatusClosed` and `NSStreamStatusError`. I figure calling `exhausted` again shouldn't be problematic. The behavior of returning 0 and leaving the status in `NSStreamStatusAtEnd` will be the same, it'll just attempt the read again first. And if a user is continuing to call `read` after reaching this state, it's probably intentional. [Apple's docs](https://developer.apple.com/documentation/foundation/nsstreamstatus/nsstreamstatusatend) aren't clear about whether `NSStreamStatusAtEnd` can be temporary or should be considered terminal.

core/apple/src/SourcesApple.kt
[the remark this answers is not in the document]
jeffdgr8: Fixed: https://github.com/Kotlin/kotlinx-io/pull/174/commits/ed5902ec5edf09000c06f7c0609b1ecca519295b

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,166 @@
...
+        status = NSStreamStatusReading
+        try {
+            if (source.exhausted()) {
+                status = NSStreamStatusAtEnd
+                return 0
+            }
+            val toRead = minOf(maxLength.toLong(), source.buffer.size, Int.MAX_VALUE.toLong()).toInt()
+            val read = source.buffer.readAtMostTo(buffer, toRead).convert<NSInteger>()
+            status = if (source.exhausted()) NSStreamStatusAtEnd else NSStreamStatusOpen
```
fzhinkin, line 74: It might be a problem: suppose we have a pair of sockets, client sends some data, we process it and respond, and there will be no more data from the client until the server respond. Here, we'll fetch all the data from a network buffer using `readAtMostTo` and then will call `exhausted` that will hang until connection timed out.
I think it's better to revert a commit changing that behavior and set the status to open after performing a read. The next read will change it to at-end if there are no more data.

jeffdgr8: Ok, I was just comparing to the behavior of `NSInputStream` and it does have the status set to `NSStreamStatusAtEnd` after the last read, at least for the `NSData` implementation. I can see how that doesn't necessarily work for all `Source` implementations though. I'll revert this.

fzhinkin:
@jeffdgr8 thank you for fixing all the issues!
Right now there are only two questions left:
- one with `exhausted` called in read (https://github.com/Kotlin/kotlinx-io/pull/174#discussion_r1282227020)
- one with inheritance from particular stream's subclass (https://github.com/Kotlin/kotlinx-io/pull/174#discussion_r1281057762)

I'll try to figure out if anything could be done with the latter (and if there's nothing we can do - I'm fine with how it's done right now).
Otherwise, the change looks good.

core/apple/src/SinksApple.kt
jeffdgr8: Related issues:

https://youtrack.jetbrains.com/issue/KT-47992
https://github.com/JetBrains/kotlin-native/issues/2010

Looks like there is a workaround for disabling in custom .def interop, but this won't work for the platform interop in this case.

core/apple/src/SourcesApple.kt
```diff
@@ -0,0 +1,166 @@
...
+import platform.Foundation.*
+import platform.darwin.NSInteger
+import platform.darwin.NSUInteger
+import platform.darwin.NSUIntegerVar
+import platform.posix.uint8_tVar
+import kotlin.native.ref.WeakReference
+
+/**
+ * Returns an input stream that reads from this source. Closing the stream will also close this source.
+ *
```
fzhinkin, line 18:
```suggestion
 *
 * The stream supports both polling and run-loop scheduling,
 * please check the [Apple's documentation](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Streams/Articles/PollingVersusRunloop.html#//apple_ref/doc/uid/20002275-CJBEDDBG)
 * for information about stream events handling.
 *
 * The stream does not implement initializers ([NSInputStream.initWithURL](https://developer.apple.com/documentation/foundation/nsinputstream/1417891-initwithurl),
 * [NSInputStream.initWithData](https://developer.apple.com/documentation/foundation/nsinputstream/1412470-initwithdata),
 * [NSInputStream.initWithFileAtPath](https://developer.apple.com/documentation/foundation/nsinputstream/1408976-initwithfileatpath)),
 * their use will result in a runtime error.
 *
```

core/apple/src/SinksApple.kt
fzhinkin: In fact, I was mostly concerned about the possibility of inheriting some methods from NSInput/OutputStream subclasses instantiated for `initFromData`, `initToMemory`. But it's definitely not the case, as there were errors due to non-implemented run-loop-related methods previously.

jeffdgr8: I added the suggested doc comments. It would be strange for a consumer to call any of those `init` functions after receiving the instantiated stream object from `asNSInputStream()` or `asNSOutputStream()`. In Kotlin it already shows an error trying to call `init` functions directly, not through a constructor. I don't think I've ever seen a call to an `init` function even in Objective-C that wasn't directly after a call to `alloc`. Swift doesn't allow this either, where `init` functions are only usable to construct objects now.

Issue #164, opened 2023-07-05
Wasm support


joreilly:
Any plans to add Wasm support? cc @bashor

fzhinkin:
@joreilly Sorry for the long reply.

I'm considering supporting Wasm target at some point.
It can not be enabled by just plunking extra target into build configs, and it requires additional effort, so I'd rather postpone the support for later.

Do you already have use cases for `kotlinx-io` in Wasm targeting projects? Could you please provide more details? Thanks!

martinbonnin:
Indeed, we've had requests for Wasm already ([issue](https://github.com/apollographql/apollo-kotlin/issues/4701)).

joreilly:
btw you probably saw this already https://github.com/square/okio/issues/1203

fzhinkin:
@joreilly @martinbonnin thank you for the details! I'll investigate what needs to be done to support Wasm then.

fzhinkin:
Everything (except file systems) works fine after updating to 1.9.10.

PR #182, opened 2023-07-18
Upgrade Gradle to 8.1

fzhinkin:
Upgraded Gradle as a step towards Kotlin 1.9 (required for https://github.com/Kotlin/kotlinx-io/issues/164) and K2.

### DIFF
--- a/benchmarks/build.gradle.kts
+++ b/benchmarks/build.gradle.kts
@@ -13,7 +13,11 @@ plugins {
 }
 
 kotlin {
-    jvm()
+    jvm {
+        jvmToolchain {
+            languageVersion.set(JavaLanguageVersion.of(libs.versions.java.get()))
+        }
+    }
     // TODO: consider supporting non-host native targets.
     if (HostManager.host === KonanTarget.MACOS_X64) macosX64("native")
     if (HostManager.host === KonanTarget.MACOS_ARM64) macosArm64("native")
@@ -28,11 +32,11 @@ kotlin {
             }
         }
 
-        val jvmMain by getting {
+        named("jvmMain") {
             dependsOn(commonMain.get())
         }
 
-        val nativeMain by getting {
+        named("nativeMain") {
             dependsOn(commonMain.get())
         }
     }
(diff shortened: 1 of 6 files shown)


shanshin (approved):

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.1-bin.zip
```
whyoleg, line 3: is there a reason not to use 8.2.1?

whyoleg: BTW, I think, that updating gradle often also involve update of wrapper.jar and gradlew files (updated when running f.e. `./gradlew wrapper --gradle-version=8.2.1`)

benchmarks/build.gradle.kts
```diff
@@ -13,7 +13,11 @@ plugins {
 }
 
 kotlin {
-    jvm()
+    jvm {
```
whyoleg, line 16: starting from gradle 8 you also need to add foojay toolchain, so JDK will be automatically downloaded if not present
more info: https://github.com/gradle/foojay-toolchains

gradle/wrapper/gradle-wrapper.properties
fzhinkin: > is there a reason not to use 8.2.1?

No, it's just me lagging behind the progress. ;)

benchmarks/build.gradle.kts
fzhinkin: I'm not 100% sure if it should be enabled, filed an issue to discuss it and enable later: https://github.com/Kotlin/kotlinx-io/issues/184

whyoleg: TBH, I don't really understand, why it should not be enabled - it simplifies contributors setup, as not all of them could have JDK 8 installed, but up to you, for me it's not an issue

PR #186, opened 2023-07-19
Migrated shared build logic from buildSrc to composite builds

### LINKED ISSUE #142: Break the curse of buildSrc and migrate to a more conventional build configuration

shanshin:
The sharing code is rewritten as convention plugins and placed in the separate build.

Two plugins are allocated - for setting up the publication and for setting up the Kotlin multiplatform project.

Resolves #142


build-logic/src/main/kotlin/kotlinx/io/conventions/multiplatform-lib-conventions.gradle.kts
```diff
@@ -95,3 +99,14 @@ fun NamedDomainObjectContainer<KotlinSourceSet>.createSourceSet(
     return result
 }
 
+kotlin {
+    configureNativePlatforms()
```
whyoleg, line 103: may be include other targets configuration here?
like JVM toolchain and junit

build.gradle.kts
```diff
@@ -31,33 +20,11 @@ allprojects {
...
-        }
-
-        val javadocJar = project.configureEmptyJavadocArtifact()
-        publications.withType(MavenPublication::class).all {
-            pom.configureMavenCentralMetadata(project)
-            signPublicationIfKeyPresent(project, this)
-            artifact(javadocJar)
-        }
-    }
+    apply(plugin = "publish-conventions")
```
whyoleg, line 27: may be apply it directly in `multplatform-lib-convention`?

shanshin: I'm not sure that only multiplatform artifacts will be published in the future.

@fzhinkin, if so, then we may think about merging.

whyoleg: Yeah, I understand that.
In my own lib I use separate convention plugins for basic configuration ([like common compiler options](https://github.com/whyoleg/cryptography-kotlin/blob/main/build-logic/src/main/kotlin/buildx-multiplatform.gradle.kts)), targets configuration ([JVM](https://github.com/whyoleg/cryptography-kotlin/blob/main/build-logic/src/main/kotlin/buildx-target-jvm.gradle.kts), [JS](https://github.com/whyoleg/cryptography-kotlin/blob/main/build-logic/src/main/kotlin/buildx-target-js.gradle.kts), [combined](https://github.com/whyoleg/cryptography-kotlin/blob/main/build-logic/src/main/kotlin/buildx-target-all.gradle.kts), and etc) and publishable modules ([with documentation and publication](https://github.com/whyoleg/cryptography-kotlin/blob/main/build-logic/src/main/kotlin/buildx-multiplatform-library.gradle.kts)).
So they could be combined then in modules: [one](https://github.com/whyoleg/cryptography-kotlin/blob/main/cryptography-core/build.gradle.kts), [two](https://github.com/whyoleg/cryptography-kotlin/blob/main/cryptography-providers/webcrypto/build.gradle.kts), [three](https://github.com/whyoleg/cryptography-kotlin/blob/main/test-support/build.gradle.kts)
While it could be overhead for just 2 modules, which we currently have, it's very scalable to new modules with different targets supported and also removes usages of `subprojects` in root module

build-logic/src/main/kotlin/kotlinx/io/conventions/multiplatform-lib-conventions.gradle.kts
```diff
@@ -100,6 +106,41 @@ fun NamedDomainObjectContainer<KotlinSourceSet>.createSourceSet(
 }
 
 kotlin {
+    jvmToolchain {
+        // Version catalog cannot be accessed from convention plugin. Until this problem is solved we configure the toolchain by project extension
+        // see https://github.com/gradle/gradle/issues/17963
+        languageVersion.set(extension.javaVersion)
```
whyoleg, line 112: just FYI, there is possibility to do this even now using type unsafe API: https://docs.gradle.org/current/userguide/platforms.html#sub:type-unsafe-access-to-catalog

shanshin: Great, thanks!

build-logic/src/main/kotlin/kotlinx/io/conventions/multiplatform-lib-conventions.gradle.kts
```diff
@@ -95,3 +99,65 @@ fun NamedDomainObjectContainer<KotlinSourceSet>.createSourceSet(
...
+        browser {
+            testTask {
+                filter.setExcludePatterns("*SmokeFileTest*")
+            }
+        }
+    }
+
+
+    sourceSets {
+        named("commonTest") {
```
whyoleg, line 128: minor: AFAIK `commonTest {}` should be resolvable out of the box, isn't it?

core/build.gradle.kts
```diff
@@ -33,48 +22,19 @@ kotlin {
...
                     timeout = "300s"
                 }
             }
         }
     }
 
-    configureNativePlatforms()
     sourceSets {
-        val commonMain by getting {
+        named("commonMain") {
```
whyoleg, line 33: same here, `commonMain {}` should be resolvable

whyoleg (commented):
last minor comments 👍

build.gradle.kts
fzhinkin: > I'm not sure that only multiplatform artifacts will be published in the future.

I'm not sure either. Can easily imagine a separate module providing integration with some platform specific API or library.

build-logic/src/main/kotlin/kotlinx/io/conventions/multiplatform-lib-conventions.gradle.kts
shanshin: Yes, thanks

build.gradle.kts
shanshin: For now, we'll leave a separate plugin for publishing, and a separate one for a multiplatform project

fzhinkin (approved):
LGTM

PR #188, opened 2023-07-19
Wasm target support

fzhinkin:
Enabled Wasm target and updated Kotlin to 1.9.10. The latter is required due to a bug in Wasm backend in 1.9.0

### DIFF
--- a/core/wasm/src/SegmentPool.kt
+++ b/core/wasm/src/SegmentPool.kt
@@ -1,0 +1,17 @@
+/*
+ * Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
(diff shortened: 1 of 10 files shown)


core/wasm/src/SegmentPool.kt
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+package kotlinx.io
+
+internal actual object SegmentPool {
+    actual val MAX_SIZE: Int = 0
+
+    actual val byteCount: Int = 0
+
+    actual fun take(): Segment = Segment()
```
shanshin, line 13: No real pool in WASM?

shanshin (approved):

core/wasm/src/SegmentPool.kt
fzhinkin: The pooling is only supported for JVM target at the moment. I'm thinking of adding support for other targets later.

PR #192, opened 2023-07-31
Revised the contents of `toString`

### LINKED ISSUE #177: Revisit `toString` content
While reviewing #174 I found that current `toString` functions are not adapted after Okio -> kotlinx-io migration:
* should be `buffered`
  https://github.com/Kotlin/kotlinx-io/blob/03845eaa0f62c4d54fc5f3ea0559e0176c2c5fa5/core/common/src/RealSink.kt#L151
* should be `asOutputStream()`
  https://github.com/Kotlin/kotlinx-io/blob/03845eaa0f62c4d54fc5f3ea0559e0176c2c5fa5/core/jvm/src/SinksJvm.kt#L85
* same for sources (may be somewhere else)

shanshin:
Resolves #177

### DIFF
--- a/core/common/src/RealSink.kt
+++ b/core/common/src/RealSink.kt
@@ -148,7 +148,7 @@ internal class RealSink(
         if (thrown != null) throw thrown
     }
 
-    override fun toString() = "buffer($sink)"
+    override fun toString() = "buffered($sink)"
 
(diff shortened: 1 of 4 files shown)


fzhinkin (approved):
LGTM

PR #193, opened 2023-08-01
Added API for reading and writing of floating point numbers

### LINKED ISSUE #167: Missing API: reading and writing of floating point numbers
Issue is present in Okio: https://github.com/square/okio/issues/202, but it is not resolved.

shanshin:
Implemented extension functions for reading and writing values with types Float and Double

Resolves #167

### DIFF
--- a/core/common/src/Sinks.kt
+++ b/core/common/src/Sinks.kt
@@ -281,6 +281,8 @@ public fun Sink.writeULongLe(long: ULong) {
  *
  * To obtain a bit representation, the [Float.toBits] function is used.
  *
+ * Should be used with care when working with special values (like `NaN`) as bit patterns obtained for [Float.NaN] may vary depending on a platform.
+ *
  * @param float the floating point number to be written.
  *
  * @throws IllegalStateException when the sink is closed.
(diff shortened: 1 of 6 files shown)


core/common/src/Sinks.kt
```diff
@@ -275,6 +275,70 @@ public fun Sink.writeULongLe(long: ULong) {
     writeLongLe(long.toLong())
 }
 
+/**
+ * Writes four bytes of a bit representation of [float], in the big-endian order, to this sink.
+ * Bit representation of the [float] corresponds to the IEEE 754 floating-point "single format" bit layout.
+ *
+ * To obtain a bit representation, the [Float.toBits] function is used.
+ *
```
fzhinkin, line 283: Does it make sense to mention that these functions should be used with care when working with special values (`NaN` particularly) as bit patterns obtained for `Float.NaN` may vary depending on a platform?

fzhinkin (approved):
LGTM

elect86:
I'd like this on `BufferedSource` as well

fzhinkin:
@elect86 there's no such entity as `BufferedSource` in `kotlinx-io`, `Source` is already buffered.

PR #194, opened 2023-08-02
Added ByteString benchmarks

### LINKED ISSUE #153: Extend benchmarking suite with ByteString-related benchmarks

fzhinkin:
Added benchmarks covering the following functions:
- indexOf
- lastIndexOf
- startsWith
- endsWith
- hashCode
- compareTo

Fixes #153

### DIFF
--- a/benchmarks/src/commonMain/kotlin/BufferOps.kt
+++ b/benchmarks/src/commonMain/kotlin/BufferOps.kt
@@ -7,6 +7,7 @@ package kotlinx.io.benchmarks
 
 import kotlinx.benchmark.*
 import kotlinx.io.*
+import kotlinx.io.bytestring.ByteString
 
 @State(Scope.Benchmark)
 abstract class BufferRWBenchmarkBase {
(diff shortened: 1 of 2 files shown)


benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -384,3 +385,33 @@ open class BufferReadNewByteArray : BufferRWBenchmarkBase() {
...
+            if (idx % bsSize == 0) {
+                buffer.writeByte(0)
+            } else {
+                buffer.writeByte(0x42)
+            }
+        }
+    }
+
+    @Benchmark
+    fun benchmark() = buffer.indexOf(byteString)
```
shanshin, line 416: In this case, the result of `indexOf` will always be 0?

benchmarks/src/commonMain/kotlin/bytestring/ByteStringOps.kt
```diff
@@ -0,0 +1,245 @@
...
+    var params: String = "<string length>:<prefix/suffix length>:<mismatch offset, or -1>"
+
+    override fun getRawParams(): String = params
+
+    @Benchmark
+    fun benchmark() = byteString.endsWith(targetByteString)
+}
+
+@State(Scope.Benchmark)
+open class CompareBenchmark {
```
shanshin, line 191: Should we also add tests on `equals`?

benchmarks/src/commonMain/kotlin/BufferOps.kt
fzhinkin: In fact, every byte at a position whose index is a multiple of bytestring size is `0`, so the benchmark has to scan the whole buffer, every time looking for up to `length-1` bytes ahead.

benchmarks/src/commonMain/kotlin/bytestring/ByteStringOps.kt
fzhinkin: Yes, indeed. Added the benchmark.

shanshin (approved):

PR #208, opened 2023-08-09
Fix potential UB in ByteString.hashCode

### LINKED ISSUE #190: Possible UB in ByteString hash function
The ByteString hash function [caches previously calculated values](https://github.com/Kotlin/kotlinx-io/blob/1ee0a1df3673e9f610cdc4fd820344b5bc6851b7/bytestring/common/src/ByteString.kt#L107-L117). The member variable that stores the cached value is a [non-atomic `Int`](https://github.com/Kotlin/kotlinx-io/blob/1ee0a1df3673e9f610cdc4fd820344b5bc6851b7/bytestring/common/src/ByteString.kt#L72). 

I believe that kotlinx-ios hash cache logic was inspired by the [google protobuf ByteString hash implementation](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/ByteString.java#L981-L999). But that hash caching has a disclaimer that the [non-atomic `Int` backing the cache is safe because of JVM memory model semantics](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/ByteString.java#L139-L144). 

Is it safe to use a non-atomic `Int` for Kotlin native builds? Does the hash cache member variable need to be converted to an `Atomic<Int>` for non-JVM platforms?

qwwdfsad:
It has a benign JVM data-race on the hashCode field, but on K/N LLVM has no guarantees it will read default or written value (e.g. OoTA is one of the possibilities). The recommendation from K/N folks is to use @Volatile here

Fixes #190

### DIFF
--- a/bytestring/common/src/-Platform.kt
+++ b/bytestring/common/src/-Platform.kt
@@ -5,7 +5,7 @@
 package kotlinx.io.bytestring
 
 /**
- * Annotation indicating that the marked property is the subjectof benign data race.
+ * Annotation indicating that the marked property is the subject of benign data race.
  * LLVM does not support this notion, so on K/N platforms we alias it into `@Volatile` to prevent potential OoTA.
  */
 @OptionalExpectation
--- a/bytestring/common/src/-Platform.kt
+++ b/bytestring/common/src/-Platform.kt
@@ -1,0 +1,14 @@
+/*
+* Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+* Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+*/
+package kotlinx.io.bytestring
+
+/**
+ * Annotation indicating that the marked property is the subjectof benign data race.
+ * LLVM does not support this notion, so on K/N platforms we alias it into `@Volatile` to prevent potential OoTA.
+ */
+@OptionalExpectation
+@Target(AnnotationTarget.FIELD)
+@OptIn(ExperimentalMultiplatform::class)
+internal expect annotation class BenignDataRace()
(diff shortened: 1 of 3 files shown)


fzhinkin:
>  OoTA is one of the possibilities

I thought it shouldn't, but:
> If there is a race on a given memory location, loads from that location return **undef**.

https://llvm.org/docs/Atomics.html#notatomic

bytestring/common/src/ByteString.kt
```diff
@@ -69,6 +70,7 @@ public class ByteString private constructor(
     public constructor(data: ByteArray, startIndex: Int = 0, endIndex: Int = data.size) :
             this(data.copyOfRange(startIndex, endIndex), null)
 
+    @Volatile
```
fzhinkin, line 73: Does it make sense to add an expect annotation like `VolatileInNative` and actualize it with `Volatile` only in native?

Like in [this patch](https://gist.github.com/fzhinkin/5cd4a5af93c6c6611ed7d8f405a55fa7)

fzhinkin: The nicest alternative would be to somehow perform load/store w/ relaxed memory ordering, though:
https://godbolt.org/z/rEdcobsW3

bytestring/common/src/-Platform.kt
```diff
@@ -0,0 +1,14 @@
+/*
+* Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+* Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+*/
+package kotlinx.io.bytestring
+
+/**
+ * Annotation indicating that the marked property is the subjectof benign data race.
```
fzhinkin, line 8:
```suggestion
 * Annotation indicating that the marked property is the subject of benign data race.
```

fzhinkin (approved):
Besides the small typo, everything LGTM. Thanks for resolving the issue!

PR #209, opened 2023-08-10
Fix Android compatibility issues

fzhinkin:


build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-android-compat.gradle.kts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+import ru.vyarus.gradle.plugin.animalsniffer.AnimalSnifferExtension
+
+apply(plugin = "ru.vyarus.animalsniffer")
+
+project.plugins.withType(JavaPlugin::class.java) {
```
shanshin, line 10: try `plugins.withType(JavaPlugin::class.java).configureEach {`

shanshin: Is `plugins.withType` used to configure AnimalSniffer only if the java plugin is specified in the project?

fzhinkin: If I correctly understood the question, when the java plugin is not specified, AnimalSniffer plugin's configuration will not affect anything (https://github.com/xvik/gradle-animalsniffer-plugin#compatibility).

shanshin: Ok, in this case, the applying of the plugin `apply(plugin = "ru.vyarus.animalsniffer")` can also be moved inside this block.

shanshin (approved):

shanshin (approved):

Issue #206, opened 2023-08-08
Make filePath (String) available


Nailik:
Currently it's not possible to read the actual Path as Text from the Path object.

Use cases are mostly native:
- sharing file
- playing with android media player (requires uri from file path)
- etc.

fzhinkin:
I don't mind exposing the string w/ a path, but I'd also need to warn that `kotlinx.io.files` is experimental and subject to change or removal.

Nailik:
@fzhinkin well it would make it easier because you could use it when parsing json data or as i said when you want to perform native actions and use the Path class commonly. This way you could then use on android the java File class and do whatever needs to be done instead of holding String and when necessary making a Path object etc etc.

Issue #212, opened 2023-08-14
Support path operations


fzhinkin:
Extend `kotlinx.io.files.Path` with the following operations:
- concatenate two paths
- get parent path

Issue #211, opened 2023-08-14
Support some basic file-system operations for kotlinx.io.files.Path


fzhinkin:
`kotlinx.io.files.Path` was added only to provide some basic operations with files and it'll be eventually replaced with proper filesystem API.

However, until it happens, there should be a way to perform some basic filesystem operations.
Following need to be added:
- delete files
- check file/directory existence
- move files atomically
- mkdirs

fzhinkin:
Also need a way to get a path to temporary dir

Issue #183, opened 2023-07-19
createTempFile() doesn't work on Apple targets


jeffdgr8:
I [discovered](https://github.com/Kotlin/kotlinx-io/pull/174/files#r1267424842) this https://github.com/Kotlin/kotlinx-io/pull/128 workaround doesn't always work on Apple platforms. The result from `mktemp()` is often not an empty string, but rather 1 or 2 bytes of seemingly junk, non-character data. This is running on macOS 13.4.1 with targets `macosX64`, `iosX64`, `tvosX64`, and `watchosX64`. The workaround sometimes works for macOS and iOS targets, but tvOS and watchOS seem to never get an empty string result to trigger the workaround.

Also, `"$tmpDir/tmp-${rnd.nextInt()}"` puts a double `//` in the path: `...//tmp-...`.

fzhinkin:
Thanks for reporting that!

On Apple platforms we should use `NSTemporaryDirectory()` to determine the temp directory.

PR #213, opened 2023-08-17
Extend filesystem support

fzhinkin:
This PR is not about bringing full filesystems support, but it rather adds few functions to make life easier.

The change adds basic path processing functions to `Path`:
- `parent` to get the parent path, backed by `dirname` on most platforms
- `name` to get the file/directory name, backed by `basename` on most platforms
- `isAbsolute` to check if the path is relative or absolute.

Paths, unlike in Okio, are platform dependent, so `a/b\\c.txt` has different meaning on Windows and Linux.

PR also brings few functions to work with filesystem. The design is driven by Okio's FileSystem, but functionality are not even close to it.
Supported functionality:
- delete file (or empty directory)
- mkdirs
- check file existence
- get file metadata (limited to isDirectory/isFile for now)
- atomic move

`kotilnx.io.files` package remains experimental (even compared to the whole library not being stable yet).

Closes: #206 #211 #212 #183


core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,113 @@
...
+     */
+    public val temporaryDirectory: Path
+
+    /**
+     * Returns `true` if there is a file system entity corresponding to a [path],
+     * otherwise returns `false`.
+     *
+     * @param path the path that should be checked for existence.
+     */
+    public fun exists(path: Path): Boolean
```
qwwdfsad, line 29: Can this one throw?

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,113 @@
...
+import kotlinx.io.IOException
+import kotlinx.io.Sink
+import kotlinx.io.Source
+
+/**
+ * An interface providing basic operations on a file system.
+ *
+ * **This API is unstable and subject to change.**
+ */
+public interface FileSystem {
```
qwwdfsad, line 17: Here and  below: let's make modality `sealed` to prevent implementation of interfaces/classes that we probably may want to evolve?

core/common/src/files/FileSystem.kt
fzhinkin: It can, thanks!

core/common/src/files/FileSystem.kt
fzhinkin: Great idea, thanks!

core/common/src/files/Paths.kt
```diff
@@ -14,20 +14,109 @@ import kotlinx.io.Source
...
+
+    override fun hashCode(): Int
+
+    /**
+     * Compares two paths for equality using its string representation ([toString]).
+     */
+    override fun equals(other: Any?): Boolean
+
+    public companion object {
+        public val separator: Char
```
qwwdfsad, line 59: should be documented

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,113 @@
...
+     * @throws kotlinx.io.IOException when the creation of one of the directories fails.
+     */
+    public fun createDirectories(path: Path, mustCreate: Boolean = false)
+
+    /**
+     * Atomically renames [source] to [destination] overriding [destination] if it already exists.
+     *
+     * When the file system does not support atomic move of [source] and [destination] corresponds to different
+     * file systems and the operation could not be performed atomically,
+     * [UnsupportedOperationException] is thrown.
```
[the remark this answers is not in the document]
fzhinkin: You may always test the code in environment where both source and destination corresponds to a single filesystem and then deploy the code to environment where these two paths are backed by different filesystems and it's impossible to perform the move atomically. I thinks it's better to explicitly document such a case so that a developer will be aware and explicitly handle it in their library/app.

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Deletes [path] from a file system. If there is no file system entity
+     * represented by the [path] this method throws [kotlinx.io.IOException] when [mustExist]
+     * is `true`.
+     *
+     * @param path the path to be deleted.
+     * @param mustExist the flag indicating whether missing [path] is an error, `true` by default.
+     *
+     * @throws kotlinx.io.IOException if the [path] could not be deleted.
```
fzhinkin, line 39: Need to document FileNotFoundException too

core/common/src/files/FileSystem.kt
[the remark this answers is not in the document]
qwwdfsad: Sure.
In the future, we _maybe_ can sweeten the pill here: e.g. call it `tryAtomicMoveOrFail` or `tryAtomicMove { fallback }` etc.
Though it requires a clear understanding of atomic move failure moves.

jeffdgr8 (commented):
Just a few doc comment fix suggestions. Looking forward to the next release with these features.

Also, if you feel inclined to fix, I noticed `utilApple.kt` got an extra redundant `ExperimentalForeignApi` `OptIn` added:
```
@file:OptIn(UnsafeNumber::class, ExperimentalForeignApi::class, ExperimentalForeignApi::class)
```

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,135 @@
...
+         * An instance of [FileSystem] representing a default system-wide file system.
+         */
+        public val System: FileSystem = SystemFileSystem
+    }
+}
+
+internal abstract class SystemFileSystemImpl : FileSystem {
+    override fun read(path: Path): Source = path.source()
+
+    override fun write(path: Path): Sink = path.sink()
```
jeffdgr8, line 125:
```suggestion
    @Suppress("DEPRECATION")
    override fun read(path: Path): Source = path.source()

    @Suppress("DEPRECATION")
    override fun write(path: Path): Sink = path.sink()
```

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+import kotlinx.io.IOException
+import kotlinx.io.Sink
+import kotlinx.io.Source
+
+/**
+ * An interface providing basic operations on a file system.
+ *
+ * **This API is unstable and subject to change.**
+ */
+public sealed interface FileSystem {
```
qwwdfsad, line 17: Could you please elaborate on why this is an interface and not a class?

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Path to a directory suitable for creating temporary files.
+     */
+    public val temporaryDirectory: Path
```
qwwdfsad, line 21: I wonder if this is a property of the file system or a system in general

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+     * otherwise returns `false`.
+     *
+     * @param path the path that should be checked for existence.
+     *
+     * @throws kotlinx.io.IOException when the attempt to check the existence of the [path] failed.
+     */
+    public fun exists(path: Path): Boolean
+
+    /**
+     * Deletes [path] from a file system. If there is no file system entity
```
qwwdfsad, line 34: The doc is a bit confusing from the user's perspective.
This function deletes a file by the given path (and/or directory?), not the path itself

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+     * Returns [Source] to read from a file represented by the [path].
+     *
+     * How a source will read the data is implementation-specific and failures caused
+     * by the missing file or, for example, lack of permissions may not be reported immediately,
+     * but postponed until the source will try to fetch data.
+     *
+     * @param path the path to read from.
+     *
+     * @throws kotlinx.io.files.FileNotFoundException when the file does not exist.
+     * @throws kotlinx.io.IOException when it's not possible to open the file for reading.
```
qwwdfsad, line 89: Could you please elaborate on what exception will be thrown if the path points to a directory?

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+    @Suppress("DEPRECATION")
+    override fun read(path: Path): Source = path.source()
+
+    @Suppress("DEPRECATION")
+    override fun write(path: Path): Sink = path.sink()
+}
+
+internal expect val SystemFileSystem: FileSystem
+
+public class FileMetadata(
```
qwwdfsad, line 132: This class is missing a root-level KDoc

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+     * Flag indicating that the metadata was retrieved for a directory.
+     */
+    public val isDirectory: Boolean = false,
+    /**
+     * File size. Defined only for regular files, for other filesystem entities it will be `-1`.
+     */
+    public val size: Long = 0L
+)
+
+public expect class FileNotFoundException(message: String?) : IOException
```
qwwdfsad, line 147: Ditto regarding KDoc

core/common/test/files/SmokeFileTest.kt
```diff
@@ -6,51 +6,255 @@
...
+        assertEquals("hello", Path("hello").name)
+        assertEquals("", Path(Path.separator.toString()).name)
+        assertEquals(".", Path(".").name)
+        assertEquals("..", Path("..").name)
+        assertEquals("hello.txt", Path("base", "hello.txt").name)
+        assertEquals("dir", Path("dir${Path.separator}").name)
+    }
+
+    @Test
+    fun isAbsolute() {
```
qwwdfsad, line 198: Could you please elaborate on why some tests are named `testFoo`, while the others are just `foo`?

No strong opinions here

qwwdfsad (changes requested):
Nice work!

Please also update README overview/links

core/common/src/files/FileSystem.kt
fzhinkin: There should be a way to implement own filesystems for testing, at least. Of course, now it's impossible due to `sealed`, but at some point, we can open it.
There's not so much to reuse between different implementations, so I decided to abstain from making it an abstract class.

core/common/src/files/FileSystem.kt
fzhinkin: The current implementation is not good at distinguishing underlying filesystems (it doesn't see the difference between `/tmp` mounted to tmpfs and `/home/usrename` mounted, let's say, to zfs).
So given then `FileSystem.System` instance, it's `temporaryDirectory` is a system-wide temporary directory.

On the other hand, it there will be a FileSystem representing some network smb share or a ssh-fs, then I would expect the different temp directory for it.

core/common/src/files/FileSystem.kt
fzhinkin: Trying to cut edges with "the package is more experimental than the whole library" in mind, sorry. Will add the docs.

core/common/src/files/FileSystem.kt
JakeWharton: Is there plan to support file systems projected over arbitrary sources of data? Like an AWS S3 bucket, an Android APK's assets directory, a zip file, or a zip file within another zip file? I don't think a temporary directory necessarily makes sense at that level.

core/common/test/files/SmokeFileTest.kt
fzhinkin: Just a lack of strong opinion on test method names, I guess. I'll use `foo` instead of `testFoo` here, thanks for pointing to it!

core/apple/src/files/FileSystemApple.kt
```diff
@@ -0,0 +1,46 @@
...
+internal actual fun basenameImpl(path: String): String {
+    memScoped {
+        return basename(path.cstr.getPointer(this))?.toKString() ?: ""
+    }
+}
+
+internal actual fun isAbsoluteImpl(path: String): Boolean = path.startsWith('/')
+
+internal actual fun mkdirImpl(path: String) {
+    val mode: UShort = 511u
```
whyoleg, line 42: may be it will be good do describe for future, what this constant means

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+     * Returns [Sink] to write into a file represented by the [path].
+     * File will be created if it does not exist yet.
+     *
+     * How a sink will write the data is implementation-specific and failures caused,
+     * for example, by the lack of permissions may not be reported immediately,
+     * but postponed until the sink will try to store data.
+     *
+     * @throws kotlinx.io.IOException when it's not possible to open the file for writing.
+     */
+    public fun write(path: Path): Sink
```
whyoleg, line 103: I have a question regarding API around `read` and `write`.
If to look at Okio they have 2 API:
* `source(Path)` and `sink(Path)` - to just obtain `Source/Sink` and use it somewhere later (result is **not** buffered source/sink)
* `read(Path)` and `write(Path)` - to use `Source/Sink` directly in place writing or reading data (lambda provides buffered source/sink

Here, we have some mix of both naming and functionality and I found current kotlinx-io way a bit confusing because while method is called `read` we don't actually perform reading yet but just `provide source for reading`. In this case naming methods `source/sink` instead of `read/write` shows intent on what we want to do better.

Also I don't really know, but may be it will be better to return `Raw*` variants here, instead of already buffered.
But that's also more a question on conventions/defaults, not only kotlinx-io but also other dependent libraries, should provide.
Should `RawSource/RawSink` should be just mostly implementation detail and buffered variants should be provided to user, or better to expose `Raw*` variants?

core/common/src/files/Paths.kt
```diff
@@ -14,20 +14,107 @@ import kotlinx.io.Source
...
+    level = DeprecationLevel.WARNING
+)
 public expect fun Path.source(): Source
 
 /**
  * Returns [Sink] for the given path, creates file if it doesn't exist, throws if it's a directory,
  * overwrites contents.
  */
+@Deprecated(
+    message = "Use FileSystem.sink instead",
```
whyoleg, line 113: there is no `FileSystem.sink` now, but `FileSystem.write` (as in ReplaceWith)

core/js/src/files/FileSystemJs.kt
```diff
@@ -0,0 +1,117 @@
...
+        return try {
+            js("require('os')")
+        } catch (t: Throwable) {
+            null
+        }
+    }
+
+private class JsFileSystem : SystemFileSystemImpl() {
+    companion object {
+        val Instance = JsFileSystem()
```
whyoleg, line 30: why can't we just make `JsFileSystem` an object?

core/js/src/files/FileSystemJs.kt
```diff
@@ -0,0 +1,117 @@
...
+                if (isFile) (stat.size as Int).toLong() else -1L
+            )
+        } catch (t: Throwable) {
+            if (exists(path)) throw IOException("Stat failed for $path", t)
+            return null
+        }
+    }
+}
+
+internal actual val SystemFileSystem: FileSystem = JsFileSystem.Instance
```
whyoleg, line 113: minor: better to use getter to avoid additional backing field (not sure if it's matter for JS)

core/jvm/src/files/FileSystemJvm.kt
```diff
@@ -0,0 +1,87 @@
...
+            override fun move(source: Path, destination: Path) {
+                throw UnsupportedOperationException("Atomic move not supported")
+            }
+        }
+    }
+}
+
+private class JvmFileSystem : SystemFileSystemImpl() {
+    companion object {
+        val Instance = JvmFileSystem()
```
whyoleg, line 46: same here as for `JsFileSystem`

whyoleg (commented):
Nice!
Some minor comments. The only major thing for me is naming of `FileSytem.read/write` methods

core/common/src/files/FileSystem.kt
fzhinkin: Made it static and renamed to `SystemTemporaryDirectory`

core/common/src/files/FileSystem.kt
fzhinkin: Fixed

core/common/src/files/FileSystem.kt
fzhinkin: Fixed

core/common/src/files/FileSystem.kt
fzhinkin: Fixed

core/common/src/files/FileSystem.kt
fzhinkin: Fixed

core/common/test/files/SmokeFileTest.kt
fzhinkin: Fixed

core/apple/test/NSInputStreamSourceTest.kt
```diff
@@ -28,9 +26,7 @@ class NSInputStreamSourceTest {
     @OptIn(ExperimentalStdlibApi::class)
     @Test
     fun nsInputStreamSourceFromFile() {
-        // can be replaced with createTempFile() when #183 is fixed
-        // https://github.com/Kotlin/kotlinx-io/issues/183
-        val file = "${NSTemporaryDirectory()}${NSUUID().UUIDString()}"
+        val file = tempFileName()
```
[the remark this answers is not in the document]
fzhinkin: Yes it is

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+     * Permissions for created directories are platform-specific.
+     *
+     * @param path the path to be created.
+     * @param mustCreate the flag indicating that existence of [path] should be treated as an error,
+     * by default it is `false`.
+     *
+     * @throws kotlinx.io.IOException when [path] already exists and [mustCreate] is `true`.
+     * @throws kotlinx.io.IOException when the creation of one of the directories fails.
+     */
+    public fun createDirectories(path: Path, mustCreate: Boolean = false)
```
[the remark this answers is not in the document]
fzhinkin: I think it's fine to abstain from reporting an error in that case as the path already exists. Not sure if we need to be more strict about what this path is actually pointing to.

core/common/src/files/Paths.kt
```diff
@@ -14,20 +14,107 @@ import kotlinx.io.Source
...
+     */
+    override fun equals(other: Any?): Boolean
+
+    public companion object {
+        /**
+         * A platform-specific character separating parts of the path.
+         *
+         * For example, it's usually `/` on Unix and `\` on Windows.
+         */
+        public val separator: Char
```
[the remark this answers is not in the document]
fzhinkin: Currently, Paths are paths in a default system's filesystem. It's unclear how to treat paths for different filesystems uniformly.

I can move the separator to FileSystem, but then some operations on Path, like `get parent` should be moved there too.

core/common/test/util.kt
```diff
@@ -39,7 +39,7 @@ fun assertNoEmptySegments(buffer: Buffer) {
     assertTrue(segmentSizes(buffer).all { it != 0 }, "Expected all segments to be non-empty")
 }
 
-expect fun createTempFile(): String
+expect fun tempFileName(): String
```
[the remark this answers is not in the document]
fzhinkin: In general, it's unsafe to use functions like `tempFileName` as there's a chance that someone will create a file with the same name in between `tempFileName` call and actual file creation.
But for tests it's fine (for me).

I'll delete `deleteFile`, thanks for pointing out!

core/js/src/files/FileSystemJs.kt
fzhinkin: It can for sure, thanks!

core/native/src/files/PathsNative.kt
```diff
@@ -18,13 +18,61 @@ import platform.posix.*
...
+
+        return path == other.path
+    }
+
+    actual override fun hashCode(): Int {
+        return path.hashCode()
+    }
+
+    public actual companion object {
+        public actual val separator: Char = '/'
```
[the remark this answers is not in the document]
fzhinkin: Windows supports both `/` and `\` as path separators. As `mingw` brings some GNU goodies to Windows, I decided to stick with `/`.

If there's a reason to choose `\` instead of `/` (like there are cases when Windows can't correctly handle it), I would definitely change the separator to `\`.

core/common/src/files/FileSystem.kt
fzhinkin: I agree that `sink/source` are better name in this context, thanks!

And it's better to return `Raw*` variants as they could be then wrapped into other decorators (to compress/decompress data, for example) and having a buffered sink/source will only add an overhead

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+     * @param path the path to read from.
+     *
+     * @throws kotlinx.io.files.FileNotFoundException when the file does not exist.
+     * @throws kotlinx.io.IOException when it's not possible to open the file for reading.
+     */
+    public fun read(path: Path): Source
+
+    /**
+     * Returns [Sink] to write into a file represented by the [path].
+     * File will be created if it does not exist yet.
```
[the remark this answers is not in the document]
fzhinkin: No way, you're right. I'll add it as an option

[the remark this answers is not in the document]
fzhinkin: Supported append

core/common/src/files/FileSystem.kt
fzhinkin: Done

core/common/src/files/Paths.kt
```diff
@@ -14,20 +16,107 @@ import kotlinx.io.Source
...
  */
+@Deprecated(
+    message = "Use FileSystem.source instead",
+    replaceWith = ReplaceWith(
+        expression = "FileSystem.System.source(this).buffered()",
+        imports = arrayOf("kotlinx.io.files.FileSystem")
+    ),
+    level = DeprecationLevel.WARNING
+)
 public expect fun Path.source(): Source
```
whyoleg, line 108: minor: looks like now expect/actual is not needed here, as implementation is now the same (via FileSystem API)

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,147 @@
...
+     * @throws kotlinx.io.IOException when [path] already exists and [mustCreate] is `true`.
+     * @throws kotlinx.io.IOException when the creation of one of the directories fails.
+     */
+    public fun createDirectories(path: Path, mustCreate: Boolean = false)
+
+    /**
+     * Atomically renames [source] to [destination] overriding [destination] if it already exists.
+     *
+     * When the file system does not support atomic move of [source] and [destination] corresponds to different
+     * file systems and the operation could not be performed atomically,
```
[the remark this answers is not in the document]
fzhinkin: Move across volumes should not work, according to MS docs (to be precise, additional flag should be passed to "MoveFile" to allow copying between volumes and I'm not passing it).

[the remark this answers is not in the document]
fzhinkin: Or did you mean two separate partitions managed by the same FS (like two distinct ext3 mounts), not Windows volumes? Atomic move won't work in that case too.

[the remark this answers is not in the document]
shanshin: Yes, I meant different mounts, for example, in Windows the paths `C:\1.txt` and `D:\1.txt`, in *nix the partitions `/dev/sda1` and `/dev/sda2`

core/common/src/files/FileSystem.kt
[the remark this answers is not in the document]
shanshin: For example, I want to create `/a/b/c` directories, but there is already a `/a/b/c` file in the file system and at the same time the `mustCreate=false` flag.
On the one hand, we cannot throw path already exists error, on the other hand, I call the function of creating directories. It turns out that an error has occurred, but I will receive a message about it postponed when I try to create a new file `/a/b/c/my.txt` (this is often what`createDirectories ` is called for)

core/common/src/files/FileSystem.kt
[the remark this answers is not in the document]
fzhinkin: Yeap, the atomic move won't work for such scenarios.

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Return [FileMetadata] associated with a file or directory the [path] points to.
+     * If there is no such file or directory, or it's impossible to fetch metadata,
+     * `null` is returned.
+     *
+     * @param path the path to get the metadata for.
+     */
+    public fun metadataOrNull(path: Path): FileMetadata?
+
+    public companion object {
```
qwwdfsad, line 124: Probablu worth leaving a `/** @suppress **/` or short documentation

core/common/src/files/Paths.kt
```diff
@@ -5,29 +5,118 @@
...
+
+    /**
+     * Returns a string representation of this path.
+     *
+     * Note that the returned value will represent the same path as the value
+     * passed to [Path], but it may not be identical to it.
+     */
+    override fun toString(): String
+
+    override fun hashCode(): Int
```
qwwdfsad, line 51: Also needs suppress or documentation.
Taking into account we already have `equals` specified, `hashCode` has more or less a single way to be specified as well

core/common/src/files/Paths.kt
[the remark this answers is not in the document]
shanshin: Perhaps it is worth explicitly specifying that only file systems of the same type are supported within one instance of the application?

core/common/src/files/Paths.kt
```diff
@@ -5,29 +5,118 @@
...
+ * Returns [RawSource] for the given file or throws if path is not a file or does not exist
  */
-public expect fun Path.source(): Source
+@Deprecated(
+    message = "Use FileSystem.source instead",
+    replaceWith = ReplaceWith(
+        expression = "FileSystem.System.source(this).buffered()",
+        imports = arrayOf("kotlinx.io.files.FileSystem")
+    ),
+    level = DeprecationLevel.WARNING
```
qwwdfsad, line 104: A maintenance tip: leave `// WARNING since 0.x, ERROR since 0.x+k, HIDDEN (if needed) since 0.x+j`

Will be easier to return to this in the future

qwwdfsad (approved):
Nice!

Have a few more or less minor comments and it's good to go

core/common/src/files/Paths.kt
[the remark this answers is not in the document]
fzhinkin: I mentioned it in the kdoc, but it may be worth moving than property to a global scope (along w/ https://github.com/Kotlin/kotlinx-io/pull/213#discussion_r1321672604) and renaming it to a `SystemPathSeparator`.

core/common/src/files/FileSystem.kt
fzhinkin: Done

core/common/src/files/Paths.kt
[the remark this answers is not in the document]
fzhinkin: Done

core/common/src/files/FileSystem.kt
```diff
@@ -0,0 +1,166 @@
...
+ */
+
+package kotlinx.io.files
+
+import kotlinx.io.IOException
+import kotlinx.io.RawSink
+import kotlinx.io.RawSource
+
+/**
+ * An interface providing basic operations on a filesystem.
```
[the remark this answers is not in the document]
fzhinkin: https://github.com/Kotlin/kotlinx-io/pull/218

PR #216, opened 2023-08-24
Correctly write data into sink's internal buffer

### LINKED ISSUE #215: Sink::asNSOutputStream overwrites sink's buffer content
`Sink::asNSOutputStream` incorrectly writes data by using the segment's `pos` offset instead of `limit` but then increments the `limit`.

fzhinkin:
Fixes #215


fzhinkin:
@jeffdgr8 PTAL

jeffdgr8 (commented):
Great find. Yes, this looks correct.

PR #227, opened 2023-10-11
Correctly extract parent path on Windows

### LINKED ISSUE #221: Path.parent doesn't work with default Windows path separator
Noting your reasoning [here](https://github.com/Kotlin/kotlinx-io/pull/213#discussion_r1312867563):
> Windows supports both `/` and `\` as path separators. As mingw brings some GNU goodies to Windows, I decided to stick with `/`.
>
> If there's a reason to choose `\` instead of `/` (like there are cases when Windows can't correctly handle it), I would definitely change the separator to `\`.

If you create a `Path` with a path string that uses Windows' default path separator `\` and call `path.parent`, the [current implementation](https://github.com/Kotlin/kotlinx-io/blob/e421c04f24dd813564a5c82c0f39ddc4548b786d/core/native/src/files/PathsNative.kt#L26) will return `null` because it doesn't handle the default path separator.

I think users would expect the `SystemPathSeparator` value to return the default path separator for the platform, being `\` for mingw.

fzhinkin:
Previously, shared native `Path.parent` implementation could filter out a windows path using `\` as a separator. 

This behavior was fixed to correctly process paths using both `\` and `/` on Windows.

Fixes #221

### DIFF
--- a/core/mingw/test/files/SmokeFileTestWindows.kt
+++ b/core/mingw/test/files/SmokeFileTestWindows.kt
@@ -25,5 +25,6 @@ class SmokeFileTestWindows {
         assertNull(Path("a\\b").parent?.parent)
         assertEquals(Path("\\\\server"), Path("\\\\server\\share").parent)
         assertEquals(Path("C:\\"), Path("C:\\Program Files").parent)
+        assertEquals(Path("C:\\Program Files"), Path("C:\\Program Files/Java").parent)
     }
 }
--- a/core/mingw/test/files/SmokeFileTestWindows.kt
+++ b/core/mingw/test/files/SmokeFileTestWindows.kt
@@ -5,9 +5,7 @@
 
 package kotlinx.io.files
 
-import kotlin.test.Test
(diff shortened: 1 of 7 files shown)


core/mingw/test/files/SmokeFileTestWindows.kt
```diff
@@ -20,4 +18,13 @@ class SmokeFileTestWindows {
         assertFalse(Path("bla\\bla\\bla").isAbsolute)
         assertTrue(Path("\\\\server\\share").isAbsolute)
     }
+
+    @Test
+    fun getParent() {
+        assertNull(Path("C:\\").parent)
```
shanshin, line 24: Shouldn't we add a mixed delimiters test, like `Path("C:\\a/b")`?

fzhinkin: The last case in this test is about mixed delimiters

shanshin (approved):

PR #228, opened 2023-10-12
Support path resolution

### LINKED ISSUE #223: Support getting a canonical file Path
Getting a file's canonical path is one of the few file APIs my library tests require that `kotlinx-io` doesn't currently support.

Similar API references:

* [JVM File.getCanonicalPath](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#getCanonicalPath--)
* [Apple Foundation NSURL.canonicalPath
](https://developer.apple.com/documentation/foundation/urlresourcevalues/1779763-canonicalpath)
* [Okio FileSystem.canonicalize](https://square.github.io/okio/3.x/okio/okio/okio/-file-system/canonicalize.html)

fzhinkin:
Fixes #223

### DIFF
--- a/core/common/src/files/FileSystem.kt
+++ b/core/common/src/files/FileSystem.kt
@@ -131,11 +131,11 @@ public sealed interface FileSystem {
     public fun metadataOrNull(path: Path): FileMetadata?
 
     /**
-     * Returns an absolute path to the same file or directory the [path] is pointing to
-     * where all symbolic links are solved, extra path separators and references to current (`.`) or
+     * Returns an absolute path to the same file or directory the [path] is pointing to.
+     * All symbolic links are solved, extra path separators and references to current (`.`) or
      * parent (`..`) directories are removed.
      * If the [path] is a relative path then it'll be resolved against current working directory.
-     * If there is no file or directory to which [path] point then [FileNotFoundException] will be thrown.
+     * If there is no file or directory to which the [path] is pointing to then [FileNotFoundException] will be thrown.
      *
      * @param path the path to resolve.
      * @return a resolved path.
--- a/core/common/src/files/FileSystem.kt
+++ b/core/common/src/files/FileSystem.kt
@@ -129,6 +129,19 @@ public sealed interface FileSystem {
      * @param path the path to get the metadata for.
      */
     public fun metadataOrNull(path: Path): FileMetadata?
+
+    /**
+     * Returns an absolute path to the same file or directory the [path] is pointing to
+     * where all symbolic links are solved, extra path separators and references to current (`.`) or
+     * parent (`..`) directories are removed.
+     * If the [path] is a relative path then it'll be resolved against current working directory.
+     * If there is no file or directory to which [path] point then [FileNotFoundException] will be thrown.
+     *
+     * @param path the path to resolve.
+     * @return a resolved path.
+     * @throws FileNotFoundException if there is no file or directory corresponding to the specified path.
+     */
+    public fun resolve(path: Path): Path
 }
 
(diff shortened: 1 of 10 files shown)


core/common/src/files/FileSystem.kt
```diff
@@ -129,6 +129,19 @@ public sealed interface FileSystem {
...
+    /**
+     * Returns an absolute path to the same file or directory the [path] is pointing to.
+     * All symbolic links are solved, extra path separators and references to current (`.`) or
+     * parent (`..`) directories are removed.
+     * If the [path] is a relative path then it'll be resolved against current working directory.
+     * If there is no file or directory to which the [path] is pointing to then [FileNotFoundException] will be thrown.
+     *
+     * @param path the path to resolve.
+     * @return a resolved path.
+     * @throws FileNotFoundException if there is no file or directory corresponding to the specified path.
```
shanshin, line 142: Is there a good reason for this behavior? It seems that our file API has functions that create a file by the passed `Path`, like `createDirectories`, `atomicMove`, `FileSystem#sink` - in this case, it will be impossible to use `resolve`.

core/common/test/files/SmokeFileTest.kt
```diff
@@ -336,7 +336,30 @@ class SmokeFileTest {
         }
         assertEquals("second third",
             SystemFileSystem.source(path).buffered().use { it.readString() })
+    }
+
+    @Test
+    fun resolve() {
```
shanshin, line 342: I think I need to add more tests: to check for an exception if the path does not exist, to check for an absolute path, etc.

core/common/src/files/FileSystem.kt
```diff
@@ -129,6 +129,19 @@ public sealed interface FileSystem {
...
+     */
+    public fun resolve(path: Path): Path
```
shanshin, line 144: Wouldn't overloading functions with other parameters be convenient to use? Like `String, vararg String` and `Path, vararg String`. 

In this case, we can write `SystemFileSystem.resolve(".")` instead of `SystemFileSystem.resolve(Path("."))`.

core/common/src/files/FileSystem.kt
fzhinkin: The `realpath` function (to which native targets delegate resolution) works only for paths corresponding to existing files/directories. That's the main reason.

shanshin: :(

core/common/src/files/FileSystem.kt
shanshin: Well, for uniformity with other functions, let it accept only `Path`

core/common/test/files/SmokeFileTest.kt
fzhinkin: Added the test for absolute path, all other cases are already there.

It would be nice to test symlink resolution, but we don't have an API for links yet.

core/common/test/files/SmokeFileTest.kt
```diff
@@ -340,26 +340,36 @@ class SmokeFileTest {
...
+        //   |-> a -> b
+        //   |-> c -> d
         val root = createTempPath()
         SystemFileSystem.createDirectories(Path(root, "a", "b"))
         val tgt = Path(root, "c", "d")
         SystemFileSystem.createDirectories(tgt)
 
         val src = Path(root, "a", "..", "a", ".", "b", "..", "..", "c", ".", "d")
         try {
+            // root/a/../a/./b/../../c/./d -> root/c/d
```
shanshin, line 364: 👍

shanshin (approved):

Issue #229, opened 2023-10-13
Update Kotlin to 1.9.20


fzhinkin:
Kotlin version could not be easily updated to 1.9.20 as there are a few issues that have to be solved first:
- KMP no longer applies `java` plugin when jvm-target is configured, it applies only java-base-plugin (see [KT-59595](https://youtrack.jetbrains.com/issue/KT-59595)). This affects the `animalsniffer` plugin [configuration](https://github.com/Kotlin/kotlinx-io/blob/7fb9df4a08da6e2be4b942b91975d2db61dbf434/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-android-compat.gradle.kts#L8);
- 1.9.20 requires to specify a `-Xexpect-actual-classes` flag to suppress warnings about expect/actual classes;
- it'll be [prohibited](https://youtrack.jetbrains.com/issue/KT-57614) to use default arguments in expect declarations actualized with typealises; we actively use it for exception classes;
- ~~[KT-22841](https://youtrack.jetbrains.com/issue/KT-22841) introduced a warning for non-final actual classes having different super-types compared to its expect declaration. `RawSink` interface is affected (on JVM in implements `java.io.Flushable`). For now, we can annotate the actual definition with `@AllowDifferentMembersInActual`;~~
- Wasm tests required nodejs update. At the moment, there are only canary v21 builds, but mocha doesn't work with them.

PR #230, opened 2023-10-13
Kotlin 1.9.20 update

fzhinkin:
See https://github.com/Kotlin/kotlinx-io/issues/229 for the changes description

### DIFF
--- a/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
+++ b/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
@@ -38,7 +38,7 @@ kotlin {
     }
 
     @OptIn(org.jetbrains.kotlin.gradle.targets.js.dsl.ExperimentalWasmDsl::class)
-    wasm {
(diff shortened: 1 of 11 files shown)


build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
```diff
@@ -198,5 +199,10 @@ fun androidTargets() = listOf(
 )
 
 rootProject.the<NodeJsRootExtension>().apply {
-    nodeVersion = "21.0.0"
+    nodeVersion = "21.0.0-v8-canary202310177990572111"
+    nodeDownloadBaseUrl = "https://nodejs.org/download/v8-canary"
```
lppedd, line 203: Is the canary version required for WASM support?

fzhinkin: Yes, it is: WASM requires v8 >= 11.9 and it is bundled only in canary builds.

shanshin (approved):

PR #236, opened 2023-11-23
Enable WasmWasi target

### LINKED ISSUE #234: Support Wasm Wasi target
After updating to Kotlin 1.9.20 only WasmJs target was enabled. We may need to support WasmWasi target as well.

fzhinkin:
Added WasmWasi target sharing the same source code with WasmJs target for now.

Closes #234

### DIFF
--- a/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
+++ b/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
@@ -46,6 +46,12 @@ kotlin {
         binaries.executable()
     }
 
+    @OptIn(org.jetbrains.kotlin.gradle.targets.js.dsl.ExperimentalWasmDsl::class)
+    wasmWasi {
+        nodejs()
+        binaries.executable()
+    }
+
     sourceSets {
         commonTest {
             dependencies {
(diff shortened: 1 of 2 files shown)


build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
```diff
@@ -96,6 +102,12 @@ kotlin {
         getByName("wasmJsTest") {
             kotlin.srcDir(File(File(projectDir, "wasm"), "test"))
         }
+        getByName("wasmWasiMain") {
+            kotlin.srcDir(File(File(projectDir, "wasm"), "src"))
```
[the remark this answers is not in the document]
fzhinkin: Thanks for pointing out the problem! I created a new sourceset `wasm`, included source dirs there and made both `wasmJs` and `wasmWasi` its dependants.

build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
```diff
@@ -46,6 +46,12 @@ kotlin {
         binaries.executable()
     }
 
+    @OptIn(org.jetbrains.kotlin.gradle.targets.js.dsl.ExperimentalWasmDsl::class)
+    wasmWasi {
+        nodejs()
+        binaries.executable()
```
lppedd, line 52: @fzhinkin sorry for bothering again, can I ask why you explicitly selected an executable binary?  
Wouldn't this be done by consumers of the library?

fzhinkin: @lppedd thanks for pointing to this! You're absolutely right, it should be done by consumers and there's absolutely no sense in specifying it for the library. I'll remove it.

PR #240, opened 2023-11-24
Avoid using PeekBuffer for newline searching

fzhinkin:
PeekBuffer creates a lot of troubles when using ByteBuffer, but even with a byte array, switching to an indexOf-based algorithm significantly improves performance.

### DIFF
--- a/core/common/src/Utf8.kt
+++ b/core/common/src/Utf8.kt
@@ -244,28 +244,29 @@ internal fun Buffer.readUtf8CodePoint(): Int {
  *
  * @sample kotlinx.io.samples.KotlinxIoCoreCommonSamples.readLinesSample
  */
+@OptIn(InternalIoApi::class)
 public fun Source.readLine(): String? {
     if (!request(1)) return null
 
-    val peekSource = peek()
-    var offset = 0L
-    var newlineSize = 0L
-    while (peekSource.request(1)) {
-        val b = peekSource.readByte().toInt()
-        if (b == '\n'.code) {
-            newlineSize = 1L
-            break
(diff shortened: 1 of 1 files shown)


fzhinkin:
Here are benchmarking results (I executed benchmarks in throughput mode, so higher the score, better the performance):

baseline (devel):
```Benchmark                          (length)  (minGap)  (separator)   Mode  Cnt         Score        Error  Units
Utf8LineBenchmark.benchmark              17       128           LF  thrpt    5  14393924.677 ± 633264.931  ops/s
Utf8LineBenchmark.benchmark              17       128         CRLF  thrpt    5  12899750.296 ± 327695.500  ops/s
Utf8LineStrictBenchmark.benchmark        17       128           LF  thrpt    5  13517716.542 ± 149910.502  ops/s
Utf8LineStrictBenchmark.benchmark        17       128         CRLF  thrpt    5  10519251.118 ±  74099.464  ops/s
```

new version (this branch):
```
Benchmark                          (length)  (minGap)  (separator)   Mode  Cnt         Score        Error  Units
Utf8LineBenchmark.benchmark              17       128           LF  thrpt    5  25948563.426 ± 525739.384  ops/s
Utf8LineBenchmark.benchmark              17       128         CRLF  thrpt    5  25194297.473 ± 615986.235  ops/s
Utf8LineStrictBenchmark.benchmark        17       128           LF  thrpt    5  25797103.301 ± 413829.337  ops/s
Utf8LineStrictBenchmark.benchmark        17       128         CRLF  thrpt    5  24884202.183 ± 542315.770  ops/s
```

shanshin (approved):

PR #242, opened 2023-11-28
Reimplement Source::indexOf(ByteString) without Source::peek calls

fzhinkin:
As in https://github.com/Kotlin/kotlinx-io/pull/240, reimplemented algorithm without the use of the `PeekSource`. Also added a specialized implementation for `Buffer`.

### DIFF
--- a/benchmarks/src/commonMain/kotlin/BufferOps.kt
+++ b/benchmarks/src/commonMain/kotlin/BufferOps.kt
@@ -393,6 +393,7 @@ open class IndexOfByteString {
 
     private var buffer = Buffer()
     private var byteString = ByteString()
+    private var source: Source = buffer
 
     @Setup
     fun setup() {
@@ -413,5 +414,5 @@ open class IndexOfByteString {
     }
 
     @Benchmark
-    fun benchmark() = buffer.indexOf(byteString)
+    fun benchmark() = source.indexOf(byteString)
 }
--- a/core/common/src/ByteStrings.kt
+++ b/core/common/src/ByteStrings.kt
@@ -102,38 +102,62 @@ public fun Source.indexOf(byteString: ByteString, startIndex: Long = 0): Long {
     }
 
     var offset = startIndex
-    val peek = peek()
-    if (!request(startIndex)) {
-        return -1L
-    }
-    peek.skip(offset)
-    var resultingIndex = -1L
-    UnsafeByteStringOperations.withByteArrayUnsafe(byteString) { data ->
-        while (!peek.exhausted()) {
-            val index = peek.indexOf(data[0])
-            if (index == -1L) {
-                return@withByteArrayUnsafe
-            }
-            offset += index
-            peek.skip(index)
-            if (!peek.request(byteString.size.toLong())) {
-                return@withByteArrayUnsafe
-            }
-
-            var matches = true
-            for (idx in data.indices) {
-                if (data[idx] != peek.buffer[idx.toLong()]) {
-                    matches = false
-                    offset++
-                    peek.skip(1)
-                    break
-                }
-            }
-            if (matches) {
-                resultingIndex = offset
-                return@withByteArrayUnsafe
-            }
+    while (request(offset + byteString.size)) {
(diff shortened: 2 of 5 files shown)


fzhinkin:
Here are benchmarking results (I executed benchmarks in throughput mode, so higher the score, better the performance):

baseline (devel):
```
Benchmark                    (params)   Mode  Cnt       Score      Error  Units
IndexOfByteString.benchmark    1024:2  thrpt    5  171907.309 ± 6796.564  ops/s
IndexOfByteString.benchmark    8192:2  thrpt    5   19421.300 ±  401.089  ops/s
IndexOfByteString.benchmark   10000:2  thrpt    5   16095.808 ±  163.348  ops/s
IndexOfByteString.benchmark   10000:8  thrpt    5    8474.794 ±   24.459  ops/s
```

new version (this branch):
```
Benchmark                    (params)   Mode  Cnt       Score       Error  Units
IndexOfByteString.benchmark    1024:2  thrpt    5  534757.836 ± 11554.425  ops/s
IndexOfByteString.benchmark    8192:2  thrpt    5   67568.970 ±   244.761  ops/s
IndexOfByteString.benchmark   10000:2  thrpt    5   51233.638 ±   365.530  ops/s
IndexOfByteString.benchmark   10000:8  thrpt    5   30259.195 ±   421.080  ops/s
```

core/common/src/ByteStrings.kt
```diff
@@ -102,38 +102,62 @@ public fun Source.indexOf(byteString: ByteString, startIndex: Long = 0): Long {
...
+                val startOffset = maxOf((startIndex - offset).toInt(), 0)
+                // Try to search the pattern within the current segment.
+                val idx = segment.indexOfBytesInbound(byteStringData, startOffset)
+                if (idx != -1) {
+                    // The offset corresponds to the segment's start, idx - to offset within the segment.
+                    return offset + idx.toLong()
+                }
+                // Try to find a pattern in all suffixes shorter than the pattern. These suffixes start
+                // in the current segment, but ends in the following segments; thus we're using outbound function.
+                for (spanningIdx in maxOf(startOffset, segment.size - byteStringData.size + 1) until segment.size) {
```
shanshin, line 147: looks like a bug, no need a loop

benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -393,6 +393,7 @@ open class IndexOfByteString {
 
     private var buffer = Buffer()
     private var byteString = ByteString()
+    private var source: Source = buffer
```
shanshin, line 396: Is this field only needed for extension function static selection?

fzhinkin: Yes, I committed that change accidentally (added it locally to ensure that Source::indexOf is also faster), but then thought that it's worth leaving it.
I think that the better solution is to introduce yet another benchmark.

core/common/src/ByteStrings.kt
fzhinkin: Indeed, it's a bug, thanks!

core/common/src/Segment.kt
```diff
@@ -210,3 +212,74 @@ internal fun Segment.indexOf(byte: Byte, startOffset: Int, endOffset: Int): Int
...
+ */
+internal fun Segment.indexOfBytesInbound(bytes: ByteArray, startOffset: Int): Int {
+    // require(startOffset in 0 until size)
+    var offset = startOffset
+    val limit = size - bytes.size + 1
+    while (offset < limit) {
+        val idx = indexOf(bytes[0], offset, limit)
+        if (idx < 0) {
+            offset++
+            continue
```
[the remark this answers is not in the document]
fzhinkin: We shouldn't, thanks!

benchmarks/src/commonMain/kotlin/BufferOps.kt
fzhinkin: Filed https://github.com/Kotlin/kotlinx-io/issues/243

shanshin (approved):

PR #244, opened 2023-11-30
Update Gradle Dokka configuration to make sure "source" button is visible in all API docs

whyoleg:
For context: https://github.com/Kotlin/kotlinx.coroutines/issues/3929 and https://github.com/Kotlin/kotlinx.coroutines/pull/3960

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -8,9 +8,9 @@ import org.jetbrains.dokka.gradle.DokkaTaskPartial
 plugins {
     id("kotlinx-io-multiplatform")
     id("kotlinx-io-publish")
+    id("kotlinx-io-dokka")
     id("kotlinx-io-android-compat")
     alias(libs.plugins.kover)
-    alias(libs.plugins.dokka)
 }
 
 kotlin {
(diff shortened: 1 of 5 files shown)


core/build.gradle.kts
```diff
@@ -8,9 +8,9 @@ import org.jetbrains.dokka.gradle.DokkaTaskPartial
 plugins {
     id("kotlinx-io-multiplatform")
     id("kotlinx-io-publish")
+    id("kotlinx-io-dokka")
```
fzhinkin, line 11: Just in case: can we also replace `alias(libs.plugins.dokka)` with `id("kotlinx-io-dokka")` in the root `build.gradle.kts`? It seems like everything works fine afterward, but maybe I'm missing something.

fzhinkin:
@whyoleg thanks for the patch! ❤️

core/build.gradle.kts
whyoleg: Not really, as root module doesn't have any sources, and so those changes have no affect for root module, also only one task from root module is executed from Dokka - `dokkaHtmlMultimodule` and it's just an aggregate task

So I would suggest to leave this as is:
* `kotlinx-io-dokka` for subprojects, where `documentation building` can be customised consistently
* plain `dokka` in root (aggregate) project, where `documentation aggregation` can be customised

It's similar to how `kover` works - there are leaf projects and there is an aggregate project (only that Dokka DSL isn't that good, yet :) )

fzhinkin: Thanks for the explanation!

fzhinkin (approved):

PR #248, opened 2023-12-08
Correctly handle paths containing whitespace character

### LINKED ISSUE #246: `Path.parent` for Native returns null for space-only parent directory
`Path.parent` utilizes `isBlank()` when checking returned value from `dirname`. This is incorrect, as a parent directory can be created as just a space.

```shell
# shell
mkdir " "
```

The following assertion should pass:
```kotlin
assertEquals(" ", Path(" /.").parent?.toString() ?: "NULL")
```

Which is the same behavior as Jvm's `java.io.File`

If `java.io.File` behavior (the underlying on Jvm) is desired to be mimicked for Native, the `dirname` check should be:

```kotlin
val parentName = dirnameImpl(path)

return when {
    parentName.isEmpty() -> null
    parentName == "." -> null
    parentName == path -> null
    else -> parentName
}
```

fzhinkin:
Fixes #246

### DIFF
--- a/core/common/src/files/Paths.kt
+++ b/core/common/src/files/Paths.kt
@@ -129,20 +129,46 @@ public fun Path.source(): Source = SystemFileSystem.source(this).buffered()
 @JvmName("sinkDeprecated")
 public fun Path.sink(): Sink = SystemFileSystem.sink(this).buffered()
 
-internal fun removeTrailingSeparators(path: String, sep: Char): String {
+internal fun removeTrailingSeparators(path: String, /* only for testing */ isWindows_: Boolean = isWindows): String {
+    if (isWindows_) {
+        // don't trim the path separator right after the drive name
+        val limit = if (path.length > 1) {
+            if (path[1] == ':') {
+                3
+            } else if (isUnc(path)) {
+                2
+            } else {
+                1
+            }
+        } else {
+            1
+        }
+        return removeTrailingSeparatorsWindws(limit, path)
+    }
+    return removeTrailingSeparatorsUnix(path)
+}
+
+private fun isUnc(path: String): Boolean {
+    if (path.length < 2) return false
+    if (path.startsWith("$WindowsPathSeparator$WindowsPathSeparator")) return true
+    if (path.startsWith("$UnixPathSeparator$UnixPathSeparator")) return true
+    return false
+}
+
+private fun removeTrailingSeparatorsUnix(path: String): String {
     var idx = path.length
-    while (idx > 1 && path[idx - 1] == sep) {
+    while (idx > 1 && path[idx - 1] == UnixPathSeparator) {
         idx--
     }
     return path.substring(0, idx)
 }
 
-internal fun removeTrailingSeparators(suffixLength: Int, path: String, sep1: Char, sep2: Char): String {
+private fun removeTrailingSeparatorsWindws(suffixLength: Int, path: String): String {
     require(suffixLength >= 1)
     var idx = path.length
     while (idx > suffixLength) {
         val c = path[idx - 1]
-        if (c != sep1 && c != sep2) break
+        if (c != WindowsPathSeparator && c != UnixPathSeparator) break
         idx--
     }
     return path.substring(0, idx)
--- a/core/common/src/files/Paths.kt
+++ b/core/common/src/files/Paths.kt
@@ -143,7 +143,7 @@ internal fun removeTrailingSeparators(path: String, /* only for testing */ isWin
         } else {
             1
         }
-        return removeTrailingSeparatorsWindws(limit, path)
+        return removeTrailingSeparatorsWindows(limit, path)
     }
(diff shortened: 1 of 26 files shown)


05nelsonm:
I was incorrect in terms of `java.io.File` parity. `null` should **not** be returned when encountering a return value of `.`

My own tests were not removing the trailing slashes when instantiating things.

`./some_path` should always return `.` for a `parent`

core/mingw/src/files/FileSystemMingw.kt
```diff
@@ -38,6 +38,11 @@ internal actual fun basenameImpl(path: String): String {
 
 internal actual fun isAbsoluteImpl(path: String): Boolean {
     if (path.startsWith(SystemPathSeparator)) return true
+    if (path.length > 1 && path[1] == ':') {
+        if (path.length == 1) return false
```
shanshin, line 42: `path.length > 1` and inside `path.length == 1` ?!

core/native/src/files/PathsNative.kt
```diff
@@ -16,17 +16,19 @@ import platform.posix.*
...
+    rawPath: String,
     @Suppress("UNUSED_PARAMETER") any: Any?
 ) {
+    internal val path = removeTrailingSeparators(rawPath)
+
     public actual val parent: Path?
         get() {
             when {
-                path.isBlank() -> return null
+                path.isEmpty() -> return null
```
shanshin, line 27: why aren't checks `!path.contains(` used like in JS?

fzhinkin: On native, that check is performed by corresponding `dirnameImpl` implementations. 
For example, https://github.com/Kotlin/kotlinx-io/blob/fe9eefc72eb2f500ab7f1f0cd5f7d1baa351d815/core/linux/src/files/FileSystemLinux.kt#L17
On all native platforms except MinGW, it checks `/` presence, but on MinGW, it also checks for `\` presence. For that reason, the check is inside `dirnameImpl` and not inside `parent`.

core/mingw/src/files/FileSystemMingw.kt
fzhinkin: Seems like I need more tests

core/js/test/files/SmokeFileTestWindowsJS.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package kotlinx.io.files
+
+import kotlinx.io.isWindows
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+
+class SmokeFileTestWindowMinGW  {
```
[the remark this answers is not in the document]
fzhinkin: Oh, sorry :(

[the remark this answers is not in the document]
fzhinkin: Added cases covering this particular bug

core/js/src/files/PathsJs.kt
```diff
@@ -84,6 +72,8 @@ public actual val SystemPathSeparator: Char by lazy {
     sep[0]
 }
 
+private const val WindowsPathSeparator: Char = '\\'
```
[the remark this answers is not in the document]
fzhinkin: Done

core/common/src/files/Paths.kt
```diff
@@ -128,3 +128,22 @@ public fun Path.source(): Source = SystemFileSystem.source(this).buffered()
 )
 @JvmName("sinkDeprecated")
 public fun Path.sink(): Sink = SystemFileSystem.sink(this).buffered()
+
+internal fun removeTrailingSeparators(path: String, sep: Char): String {
```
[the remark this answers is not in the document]
fzhinkin: Moved everything to a common source set. Now it's both easier to follow when reading the code and easier to test.

core/common/src/files/Paths.kt
```diff
@@ -129,20 +129,46 @@ public fun Path.source(): Source = SystemFileSystem.source(this).buffered()
...
+                2
+            } else {
+                1
+            }
+        } else {
+            1
+        }
+        return removeTrailingSeparatorsWindws(limit, path)
+    }
+    return removeTrailingSeparatorsUnix(path)
```
shanshin, line 148: Is it correct for all native platforms to check on `UnixPathSeparator`?
Shouldn't there be an `SystemPathSeparator` here?

fzhinkin: Only on Windows both `UnixPathSeparator` and `WindowsPathSeparator` could be used interchangeably (or at least the choice of separator in a particular path won't affect how it is treated by the FS). On all other platforms the only separator is `UnixPathSeparator`.

shanshin: but we only go into the Windows branch in tests, in other cases we always end up here

shanshin (approved):

PR #252, opened 2024-01-15
Align file sink and source creation behavior across targets

### LINKED ISSUE #251: SystemFileSystem.sink does not create empty file on NodeJs target
Reproducer (works correctly on jvm and linuxX64 but fails on jsNode):

```kotlin

package tests

import kotlinx.io.files.Path
import kotlinx.io.files.SystemFileSystem
import kotlin.test.Test
import kotlin.test.assertTrue

class ReproducerTest {

    @Test
    fun test() {
        // GIVEN
        val testFileDirPath = "/tmp/test-0001"
        val testFilePath = testFileDirPath + "/test-file.txt"

        SystemFileSystem.createDirectories(Path(testFileDirPath), false)
        assertTrue { isDirExist(testFileDirPath) }

        // WHEN creating empty file
        SystemFileSystem.sink(Path(testFilePath)).close()

        // THEN
        assertTrue { isFileExist(testFilePath) }

        // CLEANING
        SystemFileSystem.delete(Path(testFilePath), false)
    }

    fun isDirExist(directoryPath: String): Boolean {
        return SystemFileSystem.metadataOrNull(Path(directoryPath))?.isDirectory ?: false
    }

    fun isFileExist(filePath: String): Boolean {
        return SystemFileSystem.metadataOrNull(Path(filePath))?.isRegularFile ?: false
    }

}
```

fzhinkin:
For JS, unlike other targets, a file was not opened on file sink/source creation. 

For a sink, it was leading to different behavior when the sink was closed without a single byte written to it: on all targets except JS an empty file is created.

For a source, on all targets except JS, an attempt to create a source for a non-existent file is immediately reported as an error, but for JS the error won't be reported until an actual read operation is employed.

This PR aligns JS target's behavior will all other targets.

Fixes #251

### DIFF
--- a/core/js/src/files/PathsJs.kt
+++ b/core/js/src/files/PathsJs.kt
@@ -94,12 +94,12 @@ internal class FileSource(private val path: Path) : RawSource {
     private var offset = 0
     private val fd = open(path)
 
-    private fun open(path: Path): Int {
+    private fun open(path: Path): dynamic {
         if (!(fs.existsSync(path.path) as Boolean)) {
             throw FileNotFoundException("File does not exist: $path")
         }
         val fd = try {
-            fs.openSync(path.path, "r") as Int
+            fs.openSync(path.path, "r")
         } catch (e: Throwable) {
             throw IOException("Failed to open a file $path.", e)
         }
@@ -107,7 +107,6 @@ internal class FileSource(private val path: Path) : RawSource {
         return fd
     }
 
-    @OptIn(ExperimentalUnsignedTypes::class)
     override fun readAtMostTo(sink: Buffer, byteCount: Long): Long {
         check(!closed) { "Source is closed." }
         if (byteCount == 0L) {
@@ -142,10 +141,10 @@ internal class FileSink(path: Path, append: Boolean) : RawSink {
     private var closed = false
     private val fd = open(path, append)
 
-    private fun open(path: Path, append: Boolean): Int {
+    private fun open(path: Path, append: Boolean): dynamic {
         val flags = if (append) "a" else "w"
         val fd = try {
-            fs.openSync(path.path, flags) as Int
+            fs.openSync(path.path, flags)
         } catch (e: Throwable) {
             throw IOException("Failed to open a file ${path.path}.", e)
         }
--- a/core/js/src/files/PathsJs.kt
+++ b/core/js/src/files/PathsJs.kt
@@ -96,14 +96,14 @@ internal class FileSource(private val path: Path) : RawSource {
 
     private fun open(path: Path): dynamic {
         if (!(fs.existsSync(path.path) as Boolean)) {
-            throw FileNotFoundException("File does not exist: $path")
+            throw FileNotFoundException("File does not exist: ${path.path}")
         }
         val fd = try {
             fs.openSync(path.path, "r")
         } catch (e: Throwable) {
-            throw IOException("Failed to open a file $path.", e)
+            throw IOException("Failed to open a file ${path.path}.", e)
         }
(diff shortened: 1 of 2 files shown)


core/js/src/files/PathsJs.kt
```diff
@@ -92,6 +92,20 @@ internal class FileSource(private val path: Path) : RawSource {
     private var buffer: dynamic = null
```
lppedd, line 92: Just curious, why aren't you using the `Buffer` type from kotlin-wrappers for the buffer instances?

core/js/src/files/PathsJs.kt
```diff
@@ -92,6 +92,20 @@ internal class FileSource(private val path: Path) : RawSource {
...
     private var closed = false
     private var offset = 0
+    private val fd = open(path)
+
+    private fun open(path: Path): Int {
+        if (!(fs.existsSync(path.path) as Boolean)) {
+            throw FileNotFoundException("File does not exist: $path")
+        }
+        val fd = try {
+            fs.openSync(path.path, "r") as Int
```
lppedd, line 102: Generally when I know the specific numeric type, I prefer to use an `unsafeCast`.  
Casting with `as` will generate additional code paths and deeper call stack for pretty much no reasons.

core/js/src/files/PathsJs.kt
fzhinkin: It's a JS buffer returned by the [fs.readSync](https://nodejs.org/api/fs.html#fsreadsyncfd-buffer-options) and it's incompatible with `kotlinx.io.Buffer`.

The way it all works could be improved (by removing that buffer altogether and using an array from a buffer passed to the `readAtMostTo` call as a destination buffer supplied to `fs.readSync`).

core/js/src/files/PathsJs.kt
fzhinkin: Thanks for pointing to this. I don't actually care about the exact type here, so it's fine to leave it as `dynamic`.

core/js/src/files/PathsJs.kt
```diff
@@ -123,11 +133,24 @@ internal class FileSource(private val path: Path) : RawSource {
 
     override fun close() {
         closed = true
+        fs.closeSync(fd)
```
shanshin, line 136: Does calling `closeSync` won't again result in an error?

core/js/src/files/PathsJs.kt
```diff
@@ -123,11 +133,24 @@ internal class FileSource(private val path: Path) : RawSource {
...
+internal class FileSink(path: Path, append: Boolean) : RawSink {
     private var closed = false
+    private val fd = open(path, append)
+
+    private fun open(path: Path, append: Boolean): dynamic {
+        val flags = if (append) "a" else "w"
+        val fd = try {
+            fs.openSync(path.path, flags)
+        } catch (e: Throwable) {
+            throw IOException("Failed to open a file ${path.path}.", e)
```
shanshin, line 149: `${path.path}` is used here, but `$path` was above

core/js/src/files/PathsJs.kt
```diff
@@ -161,5 +179,6 @@ internal class FileSink(private val path: Path, private var append: Boolean) : R
 
     override fun close() {
         closed = true
+        fs.closeSync(fd)
```
shanshin, line 182: Same question about repeated close

core/js/src/files/PathsJs.kt
fzhinkin: It would be an error, thanks to pointing to it.

shanshin (approved):

PR #257, opened 2024-02-07
Basic filesystems support for Wasm: support WASI-based filesystem for wasmWasi

fzhinkin:


core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -5,36 +5,539 @@
...
+     * The behavior of this method differs from other platforms as the resolution
+     * may not fail if there is no filesystem-node (file, directory, symlink, etc.) corresponding
+     * to some interior path. This allows successfully resolving paths like `/a/b/c/../../d/e` when
+     * pre-opened directories are `/a/b/c` and `/a/d/e`.
+     *
+     * @param path the path to resolve.
+     * @return a resolved path.
+     * @throws FileNotFoundException if there is no file or directory corresponding to the specified path.
+     */
+    override fun resolve(path: Path): Path {
```
fzhinkin, line 229: @dkhalanskyjb (you asked, I tagged). 
Here's a hand-written canonization implementation. It's used only for Wasm-WASI.

igoriakovlev (approved):

core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -5,36 +5,539 @@
...
+
+            val ciovec = Ciovec(allocator).also {
+                it.buffer = buffer
+            }
+
+            val resultPtr = allocator.allocate(4)
+
+            while (remaining > 0) {
+                val toRead = minOf(remaining, TEMP_CIOVEC_BUFFER_LEN).toInt()
+                source.readToLinearMemory(buffer, toRead)
```
shanshin, line 383: It is not immediately clear which buffer was meant. Perhaps we should call it something else?

core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -5,36 +5,539 @@
...
+        val res = Errno(fd_sync(fd))
+        if (res != Errno.success) {
+            throw IOException("fd_sync failed: ${res.description}")
+        }
+    }
+
+    override fun close() {
+        if (!closed) {
+            closed = true
+            fd_close(fd)
```
shanshin, line 408: No exception?

core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -5,36 +5,539 @@
...
+                totalRead += readBytes
+            }
+        }
+        return totalRead
+    }
+
+    override fun close() {
+        if (!closed) {
+            closed = true
+            fd_close(fd)
```
shanshin, line 456: No exception?

core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -5,36 +5,539 @@
...
+    // parts[0] is always empty
+    for (idx in 1 until parts.size) {
+        when (val part = parts[idx]) {
+            "." -> continue
+            ".." -> constructedPath.removeLastOrNull()
+            else -> constructedPath.add(part)
+        }
+    }
+    var result = Path(UnixPathSeparator.toString())
+    for (part in constructedPath) {
```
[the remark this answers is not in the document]
fzhinkin: Oh, well, we already have `Path(root, parts)`

core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -5,36 +5,539 @@
...
+        withScopedMemoryAllocator { allocator ->
+            val (stringBuffer, stringLength) = path.path.store(allocator)
+
+            val unlinkRes = Errno(path_unlink_file(root.fd, stringBuffer.address.toInt(), stringLength))
+            if (unlinkRes == Errno.success) return
+            if (unlinkRes == Errno.noent) {
+                if (mustExist) throw FileNotFoundException("File does not exist: $path")
+                return
+            }
+            if (unlinkRes != Errno.isdir && unlinkRes != Errno.perm) {
```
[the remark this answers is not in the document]
fzhinkin: In practice, we can get a `Errno.perm` when deleting a directory (despite what is written in docs), so here we'll continue only if the resulting code is either `isdir` or `perm`.
I'll add the comment

core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -5,36 +5,539 @@
...
+                else -> throw IOException("Failed to rename $source to $destination: ${res.description}")
+            }
+        }
+    }
+
+    override fun source(path: Path): RawSource {
+        val root = PreOpens.getRoot(path)
+
+        val fd = withScopedMemoryAllocator { allocator ->
+            val fdPtr = allocator.allocate(4)
```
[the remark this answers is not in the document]
fzhinkin: For `allocate(4)` - definitely. I am not sure about Null-terminated strings. I think it should be enough to explicitly comment on why there's an extra `1` in the code for now.

core/wasmWasi/src/files/FileSystemWasm.kt
fzhinkin: We don't neither document a possible exception, nor throw it on other targets. I'd rather fix it separately along with other targets (for instance, in native we don't check a returned status).

core/wasmWasi/src/files/FileSystemWasm.kt
fzhinkin: https://github.com/Kotlin/kotlinx-io/issues/262

core/wasmWasi/test/WasiFsTest.kt
```diff
@@ -0,0 +1,96 @@
...
+            SystemFileSystem.delete(path, false)
+        }
+    }
+
+    @Test
+    fun testResolution() {
+        val resolved = SystemFileSystem.resolve(Path("/tmp/../../a/../b/../../tmp"))
+        assertEquals(Path("/tmp"), resolved)
+
+        SystemFileSystem.createDirectories(Path("/tmp/a/b/c/d/e"))
```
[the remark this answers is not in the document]
fzhinkin: Tried to create a separate test configuration, so there would be two separate templates for a test driver, but that was a mess, so I stick to simply updating the existing template.

shanshin (approved):

PR #272, opened 2024-02-29
Use default target hierarchy to setup build targets

### LINKED ISSUE #147: Use default target hierarchy to setup build targets
The build script could be cleaner with the default target hierarchy (https://kotlinlang.org/docs/whatsnew1820.html#new-approach-to-source-set-hierarchy).

And it'll be enabled by default in the future: https://youtrack.jetbrains.com/issue/KT-58676

whyoleg:
Fixes #147 

**Note: using `applyDefaultHierarchyTemplate` with extended hierarchy is `experimental` + accessors for WasmJs and WasmWasi will be available only in KGP 2.0.**

Feel free to integrate it now or wait when above points will be resolved.

Minor additional changes:
* remove `IR` in JS target, as it's default now
* minor changes to use nicer APIs for js testTasks and sourceSet dependencies
* group native targets declaration by platform, to better understand which targets are enabled - maybe it's fine to even inline `nativeTargets` function, so that all targets will be in one place


fzhinkin (approved):
LGTM. @whyoleg, thanks for taking care of it!

PR #276, opened 2024-03-05
Overload IOException constructor

### LINKED ISSUE #271: Add IOException constructor(message: Throwable?)
`IOException` lacks the secondary constructor `public constructor(message: Throwable?)`. The standard exceptions in the Kotlin stdlib and `java.io.IOException` (= the JVM actual) all have this constructor.

Would it be possible to add it?

Background: this would let us mechanically (without special casing this constructor) replace java.io.IOException with kotlinx.io.IOException as we migrate our codebase to KMP.

Goooler:
Closes #271.

### DIFF
--- a/core/common/src/-CommonPlatform.kt
+++ b/core/common/src/-CommonPlatform.kt
@@ -29,6 +29,7 @@ internal expect fun String.asUtf8ToByteArray(): ByteArray
 public expect open class IOException(message: String?, cause: Throwable?) : Exception {
     public constructor()
     public constructor(message: String?)
+    public constructor(cause: Throwable?)
 }
 
 /**
--- a/core/native/src/-NonJvmPlatform.kt
+++ b/core/native/src/-NonJvmPlatform.kt
@@ -30,7 +30,9 @@ public actual open class IOException actual constructor(
 ) : Exception(message, cause) {
     public actual constructor(message: String?) : this(message, null)
 
-    public actual constructor() : this(null)
+    public actual constructor(cause: Throwable?) : this(null, cause)
+
+    public actual constructor() : this(null, null)
 }
 
 public actual open class EOFException actual constructor(message: String?) : IOException(message) {
--- a/core/js/src/-PlatformJs.kt
+++ b/core/js/src/-PlatformJs.kt
@@ -15,7 +15,9 @@ public actual open class IOException actual constructor(
 ) : Exception(message, cause) {
     public actual constructor(message: String?) : this(message, null)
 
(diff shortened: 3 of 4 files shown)


core/native/src/-NonJvmPlatform.kt
```diff
@@ -30,7 +30,9 @@ public actual open class IOException actual constructor(
 ) : Exception(message, cause) {
     public actual constructor(message: String?) : this(message, null)
 
-    public actual constructor() : this(null)
+    public actual constructor(cause: Throwable?) : this(null, cause)
```
martinkretzschmar, line 33: The usual behavior of this exception constructor is `this(cause?.toString(), cause)`.

core/common/src/-CommonPlatform.kt
```diff
@@ -29,6 +29,7 @@ internal expect fun String.asUtf8ToByteArray(): ByteArray
 public expect open class IOException(message: String?, cause: Throwable?) : Exception {
     public constructor()
     public constructor(message: String?)
+    public constructor(cause: Throwable?)
```
fzhinkin, line 32: It does make sense to also add a constructor accepting both a message and a cause.

fzhinkin:
@Goooler thanks for opening the PR!
Please base the PR agains the `develop` branch, not the `master` (as per https://github.com/Kotlin/kotlinx-io/blob/master/CONTRIBUTING.md#submitting-changes).

fzhinkin (changes requested):

core/native/src/-NonJvmPlatform.kt
fzhinkin: I'd say it should be super(cause), and all other ctors should delegate to super ctors too.

Like here https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/wasm/src/kotlin/Exceptions.kt

core/common/src/-CommonPlatform.kt
Goooler: It's already handled by the primary constructor.

core/native/src/-NonJvmPlatform.kt
Goooler: I'll send a new PR after this gets merged.

fzhinkin (approved):

core/native/src/-NonJvmPlatform.kt
Goooler: It's in #281.

martinkretzschmar:
Thank you @Goooler!

PR #281, opened 2024-03-12
Remove exception primary constructors

Goooler:
Refs
- https://github.com/JetBrains/kotlin/blob/c961c15729f675c680d2351e8bce4ca9c9d08546/libraries/stdlib/wasm/src/kotlin/Exceptions.kt
- https://github.com/Kotlin/kotlinx-io/pull/276#discussion_r1520028548


fzhinkin (approved):
Thank you for fixing it!

Issue #241, opened 2023-11-24
Files and FileSystems support


fzhinkin:
The ability to work with files and filesystems is one of the crucial features that a programming language could provide through its standard library. Even if an application is not built around reading and writing data from storage devices, it still may need access to a filesystem to, for example, read configuration files or write logs.

A constant stream of GitHub [issues](https://github.com/Kotlin/kotlinx-io/issues?q=is%3Aissue+label%3Afs) and feedback on `kotlinx-io` suggests a demand for fully-featured multiplatform files and filesystems API.

The API exposed through `kotlinx.io.files` package was created to provide some partial file support quickly and is neither well-designed nor covers basic user needs. I'm proposing to review it, extend or redesign it if necessary.

For the purposes of this proposal, file and filesystem-related features could be split into two coarse categories: basic and extended.
Basic features include the minimal necessary features required for working with files and filesystems, something one may expect from any FS-related API. Extended features include everything else.

### Basic features

Below is the list of features to be considered as basic file and filesystem features.

#### Paths

* concatenation/joining;
* segmentation (get a parent path, get file name, file extension);
* resolution (convert relative path to absolute);
* relativization (make a path relative to a base path).

#### FileSystem features

* create/delete files;
* create/delete directories;
* list the content of a directory;
* files/directory renaming/moving (both atomic and non-atomic);
* files copying;
* directory tree traversal;
* file/directory metadata querying and update (atime/mtime/size/etc., touch);
* basic permissions support (query and update);
* symbolic links support (create, resolve);
* query and change the current working dir;
* temporary file/directory creation.

#### Files

* read (including reading from an arbitrary offset/position);
* write (including writing at an arbitrary offset/position);
* truncate/resize.

#### FileSystems

* default system filesystem;
* filesystem aimed for testing.

Most of the features listed above are presented in the vast majority of modern filesystems. However, implementation details for some of them (like file metadata or permissions) vary significantly not only between different operating systems but also within filesystems on the same OS. The way these features will be supported should be decided during the design phase.

#### Extended features

Extended features include mainly OS- or FS-specific features, niche functionality, or something that is hard to implement reliably.
The list is non-exhaustive. 

#### Paths

* an ability to manipulate with a path corresponding to a filesystem not represented in a current system (i.e. ability to explicitly process Windows paths on Unix host; see [Dart paths package docs](https://pub.dev/documentation/path/latest/) for an example).

#### FileSystem features

[...]

PR #284, opened 2024-03-18
Support directory listing

### LINKED ISSUE #222: Support listing files in a directory Path
Listing files within a directory is one of the few file APIs my library tests require that `kotlinx-io` doesn't currently support.

Similar API references:

* [JVM File.listFiles](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#listFiles--)
* [Apple Foundation NSFileManager.contentsOfDirectoryAtPath:error:](https://developer.apple.com/documentation/foundation/nsfilemanager/1414584-contentsofdirectoryatpath)
* [Okio FileSystem.list](https://square.github.io/okio/3.x/okio/okio/okio/-file-system/list.html)

fzhinkin:
It'll take a while to implement proper FS API (#241), so it does make sense to provide some API for listing directories for now.

Closes #222


core/common/src/files/FileSystem.kt
```diff
@@ -145,6 +145,20 @@ public sealed interface FileSystem {
...
+     * If path [directory] was an absolute path, a returned list will also contain absolute paths.
+     * If it was a relative path, a returned list will contain relative paths.
+     *
+     * @param directory a directory to list.
+     * @return a list of [directory]'s immediate children.
+     * @throws FileNotFoundException if [directory] does not exist.
+     * @throws IOException if [directory] points to something other than directory.
+     * @throws IOException if there was an underlying error preventing listing [directory] children.
+     */
+    public fun list(directory: Path): List<Path>
```
qwwdfsad, line 161: When looking at other ways to list the directory, folks typically avoid aggregating it in a single list, I suspect for multiple reasons:

* `opendir_at` and ability to be consistent when operating with a collection of paths (also, resolve 
* There are concerns w.r.t. performance and footprint: it is, in fact, a bulk operation that cannot be reasonably limited by the caller (and doesn't make them think to do so)
* [Kotlin-specific] "Lazy, iterable, closeable and countable sequence" is a core primitive for the filesystem, and `list` can be expressed in terms of sequence, while the opposite is not true

qwwdfsad: Probably worth considering a core methods returning a sequence and `toList` as a way to materialize it

fzhinkin: Then we need a closable sequence (which is not a thing, unlike Java's Stream).

qwwdfsad: There are other alternatives, such as `FS.forEachEntry(dir) { e -> }`.

My point was not to propose another solution but to highlight the potential pitfalls and point out that in other places, folks follow a different path [probably for a reason].
It's totally okay if you considered all of them and disregarded

core/common/test/files/SmokeFileTest.kt
```diff
@@ -443,6 +443,36 @@ class SmokeFileTest {
...
+        assertFailsWith<IOException> { SystemFileSystem.list(tmpFile) }
+
+        val dir = createTempPath().also {
+            SystemFileSystem.createDirectories(it)
+        }
+        assertEquals(emptyList(), SystemFileSystem.list(dir))
+
+        val subdir = Path(dir, "subdir").also {
+            SystemFileSystem.createDirectories(it)
+            SystemFileSystem.sink(Path(it, "file")).close()
```
shanshin, line 462: Why we create file here?

core/common/test/files/SmokeFileTest.kt
```diff
@@ -443,6 +443,36 @@ class SmokeFileTest {
...
+        }
+        assertEquals(listOf(subdir), SystemFileSystem.list(dir))
+
+        val file = Path(dir, "file").also {
+            SystemFileSystem.sink(it).close()
```
shanshin, line 467: Why not test other file types, such as symlinks?

core/common/src/files/FileSystem.kt
shanshin: It seems to me that ideally there should be something like what is in stdlib, but I'm not sure that this needs to be done right now.
Is it worth adding an API now that we know in advance that we will remove it soon?!

build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
```diff
@@ -80,6 +80,12 @@ kotlin {
                         group("androidNative")
                     }
                 }
+
+                group("nonAndroid") {
```
shanshin, line 84: Am I right that `nonAndroid` is now actually a synonym for `posix`?

core/common/test/files/SmokeFileTest.kt
fzhinkin: We don't support symlinks yet (in wasmWasi it was supported only to make other things work correctly).

core/common/test/files/SmokeFileTest.kt
fzhinkin: To ensure it won't be listed (i.e. that `list` is not recursive).

core/common/test/files/SmokeFileTest.kt
shanshin: Symlinks can't be listed from directory?

core/common/src/files/FileSystem.kt
fzhinkin: >  soon?!

It depends on how you perceive time, I'm afraid.

build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
fzhinkin: Well, the only POSIX certified target down there is macOs :)

core/common/test/files/SmokeFileTest.kt
fzhinkin: We don't have an API to create them for testing purposes. They can be listed, of course.

build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
shanshin: But for somehow we use posix API for all of them ))

shanshin: It's just that for me, JVM is also a non-Android, so this name was confused at first

fzhinkin: But Android is as POSIX as Linux

fzhinkin: Perhaps, `commonizedAsPosix`?

core/wasmWasi/test/WasiFsTest.kt
```diff
@@ -219,4 +219,14 @@ class WasiFsTest {
...
         }
     }
+
+    // https://github.com/nodejs/node/blob/6f4d6011ea1b448cf21f5d363c44e4a4c56ca34c/deps/uvwasi/src/uvwasi.c#L19
+    @Test
+    fun readdirUnsupportedOnWindows() {
+        if (!isWindows) return
+
+        assertFailsWith<IOException> {
+            SystemFileSystem.list(PreOpens.roots.first())
```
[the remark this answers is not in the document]
fzhinkin: I don't see how we can overcome this issue. Perhaps, the best thing we can do is explicitly document that directory listing does not work for wasmWasi target in conjunction with NodeJs on Windows.

build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
fzhinkin: Yet another strenuous attempt to give a better name:
- `nativeNonAndroid`
- rename `nonApple` to `nativeNonApple`

WDYT?

core/wasmWasi/test/WasiFsTest.kt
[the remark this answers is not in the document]
fzhinkin: Opened https://github.com/Kotlin/kotlinx-io/issues/312

Also, I'll document lack of support on NodeJs/Windows

core/common/test/files/SmokeFileTest.kt
shanshin: Issue https://github.com/Kotlin/kotlinx-io/issues/241

core/wasmWasi/test/WasiFsTest.kt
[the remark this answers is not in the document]
fzhinkin: Done

StefanOltmann:
Nice! :)

Issue #265, opened 2024-02-23
Kotlin/JS-modules no longer work in browser after #256


fzhinkin:
In #256 I refactored the interop with nodeJs modules providing filesystem support. If previously, only functionality related to files and filesystem didn't work in a browser (with UnsupportedOperationException being thrown), now an attempt to use anything declared in kotlinx-io-core leads to errors like this:
```
> Task :gradle-core-multiplatform:jsBrowserTest
Module not found: Error: Can't resolve 'buffer' in '/Users/Filipp.Zhinkin/Development/kotlinx-io-release-tests/build/js/packages/kotlinx-io-release-tests-gradle-core-multiplatform-test/kotlin'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "buffer": require.resolve("buffer/") }'
        - install 'buffer'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "buffer": false }
Module not found: Error: Can't resolve 'fs' in '/Users/Filipp.Zhinkin/Development/kotlinx-io-release-tests/build/js/packages/kotlinx-io-release-tests-gradle-core-multiplatform-test/kotlin'
Module not found: Error: Can't resolve 'os' in '/Users/Filipp.Zhinkin/Development/kotlinx-io-release-tests/build/js/packages/kotlinx-io-release-tests-gradle-core-multiplatform-test/kotlin'
```

It's a blocker as it makes kotlinx-io-core useless in nodeJs-less environments.

lppedd:
This really shows how much we need https://youtrack.jetbrains.com/issue/KT-47038.

This type of issue would be entirely avoidable.

joffrey-bion:
Is there any workaround for this?

joffrey-bion:
I just realized #256 was not released yet. I'm using 0.3.1 and I'm experiencing these errors already. So I'm not sure they are caused by the change you mention @fzhinkin

fzhinkin:
@joffrey-bion I checked it with 0.3.1 and got error messages, but the code using the library worked fine (modulo `UnsupportedOperationExcepion` on an attempt to use a filesystem). Could you please elaborate on what kind of error you're facing?

joffrey-bion:
[...]

PR #285, opened 2024-03-19
Trick the webpack to bypass its warnings on missing modules

fzhinkin:
Leftovers from #265

### DIFF
--- a/core/js/src/node/nodeModulesJs.kt
+++ b/core/js/src/node/nodeModulesJs.kt
@@ -7,7 +7,7 @@ package kotlinx.io.node
 
 internal actual val path: Path by lazy {
     try {
-        js("require(\"path\")")
+        js("eval('require')('path')")
     } catch (e: Throwable) {
         throw UnsupportedOperationException("Module 'path' could not be imported", e)
     }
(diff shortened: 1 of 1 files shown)


whyoleg (approved):
LGTM
But, is there a reason why Wasm and JS implementations are different?
In my experiments current approach works for both Wasm and JS, and so the implementation looks similar (except for minor differences in jsinterop for wasm vs js), so it's easier to understand the code and so could mostly copy-pasted

fzhinkin:
> But, is there a reason why Wasm and JS implementations are different?

There was a reason, but now it seems like it should be possible to unify it.

fzhinkin:
Filed https://github.com/Kotlin/kotlinx-io/issues/287 for unifying module loading

lppedd:
I've missed the discussion about this change. Why couldn't Webpack simply be configured to mute those warnings?

E.g., configure `externals`, or use the ignore plugin.

fzhinkin:
@lppedd If I understand correctly, it could be done on the library user side only, which is kinda lame. Instead, the root cause of the warning was fixed (well, not fixed, but muted).

lppedd:
@fzhinkin Thanks. I get that not displaying a warning consumer side is nice, but isn't it better to make the consumer aware of what is going on instead of hiding it? I mean, once I start working with K/JS I know that Webpack runs for `browser`, so I can fiddle with its configuration in case. This could be highlighted in the readme.

PR #286, opened 2024-03-19
Implement path normalization

### LINKED ISSUE #223: Support getting a canonical file Path

fzhinkin:
Closes #223

### DIFF
--- a/core/common/src/files/Paths.kt
+++ b/core/common/src/files/Paths.kt
@@ -41,6 +41,12 @@ public expect class Path {
     public val isAbsolute: Boolean
 
     /**
+     * Returns normalized version of this path where all `..` and `.` segments are resolved
+     * and all sequential path separators are collapsed.
+     */
+    public fun normalized(): Path
+
+    /**
      * Returns a string representation of this path.
      *
      * Note that the returned value will represent the same path as the value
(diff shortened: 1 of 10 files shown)


core/common/src/files/Paths.kt
```diff
@@ -174,3 +180,51 @@ private fun removeTrailingSeparatorsWindows(suffixLength: Int, path: String): St
     }
     return path.substring(0, idx)
 }
+
+internal fun Path.normalizedInternal(preserveDrive: Boolean, vararg separators: Char): String {
```
fzhinkin, line 184: It works incorrectly for UNC paths

fzhinkin: Will fix it

PR #295, opened 2024-04-15
Add toFile and toNioPath for Path on JVM

### LINKED ISSUE #292: Add a `Path.toFile()` converter on JVM
…and perhaps also, a `Path.toNioPath()` on JVM.

Okio has a similar [`Path.toFile()`](https://square.github.io/okio/3.x/okio/okio/okio/-path/to-file.html) method, along with a complementary [`Path.toNioPath()`](https://square.github.io/okio/3.x/okio/okio/okio/-path/to-nio-path.html).

While you could technically convert a `Path` into a string and then create a new `java.io.File` out of it, it feels like a waste when there's already a `File` object underneath that we could use. If `kotlinx-io` has a `Path.toFile()`, then `Path.toNioPath()` could also simply be implemented as `path.toFile().toPath()`, which would also return the same NIO `Path` instance every time.

Now of course, there's also the possibility that the underlying object implementing `Path` will not be a `File` anymore in the future. So the described performance win is irrelevant. However, I also think having such utility conversion methods would be convenient especially when trying to interoperate with JVM (or Android). It's useful for providing features currently present in the JVM but not in `kotlinx-io`, such as when implementing a custom `FileSystem` backed by JVM APIs, or when adding platform-dependent features. Also useful when you need a quick way to interop with some Java library that happens to require a `File` or NIO `Path`.

Parsing and manipulating paths is also one area where you're not touching the underlying file system to induce IO overhead, but when doing a lot of parsing and manipulation, you generally want that to be performant, so as to not waste precious CPU and battery life unnecessarily, especially when on Android.

Goooler:
Closes #292.

### DIFF
--- a/core/api/kotlinx-io-core.api
+++ b/core/api/kotlinx-io-core.api
@@ -239,6 +239,8 @@ public final class kotlinx/io/files/PathsJvmKt {
 	public static final fun Path (Ljava/lang/String;)Lkotlinx/io/files/Path;
 	public static final fun sink (Lkotlinx/io/files/Path;)Lkotlinx/io/Sink;
 	public static final fun source (Lkotlinx/io/files/Path;)Lkotlinx/io/Source;
+	public static final fun toFile (Lkotlinx/io/files/Path;)Ljava/io/File;
+	public static final fun toNioPath (Lkotlinx/io/files/Path;)Ljava/nio/file/Path;
 }
 
 public final class kotlinx/io/files/PathsKt {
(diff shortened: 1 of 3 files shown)


fzhinkin:
Unfortunately, such converters work correctly only for a local filesystem, and supporting them for the general case might not be feasible (see https://github.com/Kotlin/kotlinx-io/issues/292#issuecomment-2074702225).

Closing the PR for now, we can always reconsider it later.

PR #308, opened 2024-05-03
Make functions to read and write code points public

### LINKED ISSUE #307: Make functions to read and write code points public
Currently, there's no way to read a code point from a source containing UTF-8 encoded data or write a single code point to sink encoding it to UTF-8 along the way.

Corresponding functions exist but are hidden: https://github.com/Kotlin/kotlinx-io/pull/136#discussion_r1243305835

It's time to make them public under the following names:
- `Sink.writeCodePointValue(value: Int)`
- `Source.readCodePointValue(): Int`

There is no `Utf8` in the names for consistency with `readString`/`writeString`, but there is a `Value` suffix.
At some point, `CodePoint` type will be added to Kotlin Stdlib, so it'll be nice to preserve `readCointPoint`/`writeCodePoint` names for it. However, such a type most likely will have a `value: Int` property and the names mentioned above are referring to that property.

fzhinkin:
Fixes #307


fzhinkin:
KDoc preview:
- [readCodePointValue](https://teamcity.jetbrains.com/repository/download/KotlinTools_KotlinxIo_GenerateDocumentation/4603077:id/kotlinx-io-core/kotlinx.io/read-code-point-value.html)
- [writeCodePointValue](https://teamcity.jetbrains.com/repository/download/KotlinTools_KotlinxIo_GenerateDocumentation/4603077:id/kotlinx-io-core/kotlinx.io/write-code-point-value.html)

core/common/src/Utf8.kt
```diff
@@ -122,14 +122,23 @@ internal fun String.utf8Size(startIndex: Int = 0, endIndex: Int = length): Long
 /**
  * Encodes [codePoint] in UTF-8 and writes it to this sink.
  *
+ * Note that in general, a value retrieved from [Char.code] could not be written directly
+ * as it may be a part of a [surrogate pair](https://www.unicode.org/faq/utf_bom.html#utf16-2) (that could be
+ * detected using [Char.isSurrogate], or [Char.isHighSurrogate] and [Char.isLowSurrogate]).
+ * Such a pair of characters needs to be manually converted back to a single code point
+ * which then could be written to a [Sink].
+ * Without such a conversion, data written to a [Sink] will no
```
sandwwraith, line 130: Maybe `can not be converted`?

core/common/src/Utf8.kt
```diff
@@ -196,24 +205,29 @@ public fun Source.readString(byteCount: Long): String {
...
+ * Decodes a single code point value from UTF-8 code units, reading between 1 and 4 bytes as necessary.
  *
  * If this source is exhausted before a complete code point can be read, this throws an
  * [EOFException] and consumes no input.
  *
- * If this source doesn't start with a properly-encoded UTF-8 code point, this method will remove
+ * If this source doesn't start with a properly encoded UTF-8 code point, this method will remove
  * 1 or more non-UTF-8 bytes and return the replacement character (`U+fffd`). This covers encoding
- * problems (the input is not properly-encoded UTF-8), characters out of range (beyond the
+ * problems (the input is not properly encoded UTF-8), characters out of range (beyond the
```
sandwwraith, line 215: > If this source doesn't start with a properly encoded UTF-8 code point, this method will remove
1 or more non-UTF-8 bytes and return the replacement character (`U+fffd`). This covers <...> code points for UTF-16 surrogates (`U+d800`..`U+dfff`)

What this means, exactly? We can decode 1 whole code point that later forms a surrogate pair. Then in what case UTF-16 surrogates are replaced with a replacement char?

core/common/src/Utf8.kt
```diff
@@ -196,24 +205,29 @@ public fun Source.readString(byteCount: Long): String {
...
 /**
- * Removes and returns a single UTF-8 code point, reading between 1 and 4 bytes as necessary.
+ * Decodes a single code point value from UTF-8 code units, reading between 1 and 4 bytes as necessary.
  *
  * If this source is exhausted before a complete code point can be read, this throws an
  * [EOFException] and consumes no input.
  *
- * If this source doesn't start with a properly-encoded UTF-8 code point, this method will remove
+ * If this source doesn't start with a properly encoded UTF-8 code point, this method will remove
  * 1 or more non-UTF-8 bytes and return the replacement character (`U+fffd`). This covers encoding
```
sandwwraith, line 214: Given that we write `?` instead of a proper replacement char on encoding, shouldn't we do the same for decoding?

fzhinkin: Currently, such a behavior is consistent with how `readString`/`writeString` works.

Given that the problem also arose around `utf8Size`, I think it should be resolved separately to take various options and opinions into account.

core/common/src/Utf8.kt
fzhinkin: > What this means, exactly?

If a prefix of data contained in a `Source` is an ill-formed code unit sequence from the UTF-8 point of view, this method will remove these code units from the source and return the replacement code point.

If there's a well-formed code unit sequence, it will be removed from the Source and decoded. If a decoded value falls out of valid UTF-16 code point value ranges, we'll also return the replacement code point.

I think that rephrasing `If this source doesn't start with a properly encoded UTF-8 code point` to `If this source starts with an ill-formed UTF-8 code unit sequence` should help with translating the idea. WDYT?

core/common/src/Utf8.kt
```diff
@@ -196,24 +205,31 @@ public fun Source.readString(byteCount: Long): String {
...
- * 1 or more non-UTF-8 bytes and return the replacement character (`U+fffd`). This covers encoding
- * problems (the input is not properly-encoded UTF-8), characters out of range (beyond the
- * `0x10ffff` limit of Unicode), code points for UTF-16 surrogates (`U+d800`..`U+dfff`) and overlong
- * encodings (such as `0xc080` for the NUL character in modified UTF-8).
+ * If this source starts with an ill-formed UTF-8 code units sequence, this method will remove
+ * 1 or more non-UTF-8 bytes and return the replacement character (`U+fffd`).
+ *
+ * The replacement character (`U+fffd`) will be also returned if the source starts with a well-formed
+ * code units sequences, but a decoded value does not pass further validation, such as
+ * the value is of range (beyond the `0x10ffff` limit of Unicode), maps to UTF-16 surrogates (`U+d800`..`U+dfff`),
```
sandwwraith, line 218:
```suggestion
 * the value is out of range (beyond the `0x10ffff` limit of Unicode), maps to UTF-16 surrogates (`U+d800`..`U+dfff`),
```

It's better now, but maybe my Unicode knowledge still fails me: why `a decoded value does not pass further validation` if it `maps to UTF-16 surrogates ('U+d800'..'U+dfff')`?

fzhinkin: > why a decoded value does not pass further validation if it maps to UTF-16 surrogates ('U+d800'..'U+dfff')

3. Conformance / § 3.9 Unicode Encoding Forms / D92
```
Because surrogate code points are not Unicode scalar values, 
any UTF-8 byte sequence that would otherwise map to code points U+D800..U+DFFF is ill-formed.
```

https://www.unicode.org/versions/Unicode15.1.0/ch03.pdf

sandwwraith: Ah, I guess the meaning is that 'byte sequence should form a whole code point, not a half of it' and 'valid utf-16 is not always a valid utf-8'.

sandwwraith (approved):

core/common/src/Utf8.kt
fzhinkin: > byte sequence should form a whole code point, not a half of it

That's a funny part:
- Unicode code point is an integer in [0, 0x1FFFF] range (D10, D9 from 3.4 Characters and Encoding)
- Unicode scalar values are all code points except surrogates (D76 from 3.9 Unicode Encoding Forms)
- Surrogates should not be interpreted as abstract characters and are subject to restrictions in interchanges as per 2.4 Code Points and Characters.

But the abovementioned D92 simply restricts encoding surrogates using UTF-8, despite surrogates being (almost) regular Unicode code points.

core/common/src/Utf8.kt
```diff
@@ -122,14 +122,23 @@ internal fun String.utf8Size(startIndex: Int = 0, endIndex: Int = length): Long
...
  * Encodes [codePoint] in UTF-8 and writes it to this sink.
  *
+ * Note that in general, a value retrieved from [Char.code] could not be written directly
+ * as it may be a part of a [surrogate pair](https://www.unicode.org/faq/utf_bom.html#utf16-2) (that could be
+ * detected using [Char.isSurrogate], or [Char.isHighSurrogate] and [Char.isLowSurrogate]).
+ * Such a pair of characters needs to be manually converted back to a single code point
+ * which then could be written to a [Sink].
+ * Without such a conversion, data written to a [Sink] can not be converted back
+ * to a string from which a surrogate pair was retrieved.
+ *
```
shanshin, line 132: The documentation does not say anything about what will be written if we pass a code point with the surrogate value (`U+d800`..`U+dfff`).
E.g. on the JVM will be written single byte with value 63 (`?`)

Also we should add test on it

core/common/src/Utf8.kt
```diff
@@ -196,24 +205,31 @@ public fun Source.readString(byteCount: Long): String {
...
  *
- * If this source doesn't start with a properly-encoded UTF-8 code point, this method will remove
- * 1 or more non-UTF-8 bytes and return the replacement character (`U+fffd`). This covers encoding
- * problems (the input is not properly-encoded UTF-8), characters out of range (beyond the
- * `0x10ffff` limit of Unicode), code points for UTF-16 surrogates (`U+d800`..`U+dfff`) and overlong
- * encodings (such as `0xc080` for the NUL character in modified UTF-8).
+ * If this source starts with an ill-formed UTF-8 code units sequence, this method will remove
+ * 1 or more non-UTF-8 bytes and return the replacement character (`U+fffd`).
+ *
+ * The replacement character (`U+fffd`) will be also returned if the source starts with a well-formed
```
shanshin, line 216: Should we add test on ill-formed UTF-8?

Two tests if ill-formed at start or at the end of source

core/common/src/Utf8.kt
fzhinkin: Currently, we always writing '?'. `Utf8Test::writeSurrogateCodePoint` covers that scenario.

core/common/src/Utf8.kt
fzhinkin: We have some tests on ill-formed UTF-8, and there's a test dedicated to surrogates (`Utf8Test::readSurrogateCodePoint`)

core/common/src/Utf8.kt
shanshin: But this is not mentioned in the KDoc

fzhinkin: Thanks for pointing to that! Opened https://github.com/Kotlin/kotlinx-io/pull/314

core/common/src/Utf8.kt
```diff
@@ -122,14 +122,23 @@ internal fun String.utf8Size(startIndex: Int = 0, endIndex: Int = length): Long
...
  * @throws IllegalStateException when the sink is closed.
  *
- * @sample kotlinx.io.samples.KotlinxIoCoreCommonSamples.utf8CodePointSample
+ * @sample kotlinx.io.samples.KotlinxIoCoreCommonSamples.writeUtf8CodePointSample
+ * @sample kotlinx.io.samples.KotlinxIoCoreCommonSamples.writeSurrogatePair
  */
 @OptIn(DelicateIoApi::class)
-internal fun Sink.writeUtf8CodePoint(codePoint: Int): Unit =
+public fun Sink.writeCodePointValue(codePoint: Int): Unit =
     writeToInternalBuffer { it.commonWriteUtf8CodePoint(codePoint) }
```
ilya-g, line 142: Does it handle negative values of `codePoint`?

fzhinkin: It's not. Opened https://github.com/Kotlin/kotlinx-io/pull/317

PR #314, opened 2024-05-07
Improve writeCodePointValue KDoc

fzhinkin:
Follow up to https://github.com/Kotlin/kotlinx-io/pull/308

### DIFF
--- a/core/common/src/Utf8.kt
+++ b/core/common/src/Utf8.kt
@@ -130,6 +130,9 @@ internal fun String.utf8Size(startIndex: Int = 0, endIndex: Int = length): Long 
  * Without such a conversion, data written to a [Sink] can not be converted back
  * to a string from which a surrogate pair was retrieved.
  *
+ * More specifically, all code points mapping to UTF-16 surrogates (`U+d800`..`U+dfff`)
+ * will be written as `?` characters (`U+0063`).
+ *
  * @param codePoint the codePoint to be written.
  *
  * @throws IllegalStateException when the sink is closed.


shanshin (approved):

core/common/src/Utf8.kt
```diff
@@ -130,6 +130,9 @@ internal fun String.utf8Size(startIndex: Int = 0, endIndex: Int = length): Long
  * Without such a conversion, data written to a [Sink] can not be converted back
  * to a string from which a surrogate pair was retrieved.
  *
+ * More specifically, all code points mapping to UTF-16 surrogates (`U+d800`..`U+dfff`)
+ * will be written as `?` characters (`U+0063`).
```
ilya-g, line 134: Is this JVM-only behavior or common one?

core/common/src/Utf8.kt
```diff
@@ -130,6 +130,9 @@ internal fun String.utf8Size(startIndex: Int = 0, endIndex: Int = length): Long
...
+ *
  * @param codePoint the codePoint to be written.
  *
  * @throws IllegalStateException when the sink is closed.
```
ilya-g, line 138: Makes sense to document that IllegalArgumentException can be thrown.

core/common/src/Utf8.kt
fzhinkin: Currently, it's a common one

core/common/src/Utf8.kt
fzhinkin: https://github.com/Kotlin/kotlinx-io/pull/317

core/common/src/Utf8.kt
ilya-g: This makes it inconsistent with `String.encodeToByteArray` where the replacement is unspecified, but different.

fzhinkin: `encodeToByteArray` call on a string with a low surrogate only results in a byte array containing `?`: https://pl.kotl.in/uFX3YULin

PR #318, opened 2024-05-08
Add Sink.writeString(CharSequence)

### LINKED ISSUE #261: Missing API: `Sink.write(CharSequence)`
At the moment, when receiving data from a mutable `CharSequence`, one has to use `sink.writeString(chars.toString())`, which may be inefficient (copying data to create a `String`, just to then encode it again as UTF-8 bytes in the buffer).

It would be nice to have a `Sink.writeUtf8(CharSequence)` extension so we can efficiently transfer characters from a `CharSequence` into a UTF-8-encoded `Buffer`.

fzhinkin:
Fixes #261

### DIFF
--- a/core/common/src/Utf8.kt
+++ b/core/common/src/Utf8.kt
@@ -162,8 +162,31 @@ public fun Sink.writeCodePointValue(codePoint: Int): Unit =
  * @sample kotlinx.io.samples.KotlinxIoCoreCommonSamples.writeUtf8Sample
  */
 @OptIn(DelicateIoApi::class)
-public fun Sink.writeString(string: String, startIndex: Int = 0, endIndex: Int = string.length): Unit =
-    writeToInternalBuffer { it.commonWriteUtf8(string, startIndex, endIndex) }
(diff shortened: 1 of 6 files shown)


core/common/src/Utf8.kt
```diff
@@ -162,8 +162,31 @@ public fun Sink.writeCodePointValue(codePoint: Int): Unit =
...
+
+    writeToInternalBuffer { it.commonWriteUtf8(startIndex, endIndex, string::get) }
+}
+
+/**
+ * Encodes the characters at [startIndex] up to [endIndex] from [chars] in UTF-8 and writes it to this sink.
+ *
+ * @param chars the string to be encoded.
+ * @param startIndex the index (inclusive) of the first character to encode, 0 by default.
+ * @param endIndex the index (exclusive) of a character past to a last character to encode, `string.length` by default.
```
shanshin, line 176:
```suggestion
 * @param endIndex the index (exclusive) of a character past to a last character to encode, `chars.length` by default.
```

shanshin (approved):

PR #320, opened 2024-05-08
Add a benchmark on individual Unicode codepoints reading/writing

fzhinkin:


benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -415,3 +416,67 @@ open class IndexOfByteString {
...
     fun benchmark() = buffer.indexOf(byteString)
 }
+
+@State(Scope.Benchmark)
+open class Utf8CodePointsBenchmark : BufferRWBenchmarkBase() {
+    private val codePointsCount = 128
+
+    // Encoding names follow naming from Utf8StringBenchmark
+    @Param("ascii", "utf8", "sparse", "2bytes", "3bytes", "4bytes", "bad")
+    var encoding: String = "ascii"
```
shanshin, line 426: Maybe it's better to use enum in such cases?

benchmarks/src/commonMain/kotlin/BufferOps.kt
```diff
@@ -415,3 +416,67 @@ open class IndexOfByteString {
...
+    @Setup
+    fun fillCodePointsArray() {
+        fun IntArray.fill(generator: () -> Int) {
+            for (idx in this.indices) {
+                this[idx] = generator()
+            }
+        }
+
+        when (encoding) {
+            "ascii" -> codePoints.fill { Random.nextInt(' '.code, '~'.code) }
```
shanshin, line 444: Why are codes less than 0x20 ignored?)

shanshin (approved):

benchmarks/src/commonMain/kotlin/BufferOps.kt
fzhinkin: Just a personal preference :)

benchmarks/src/commonMain/kotlin/BufferOps.kt
fzhinkin: I decided to use the same values as for the old read/writeString benchmark. I can replace the parameter with enum, but then the old value need to be replaced (especially `2|3|4bytes`), and that would complicate results comparison against older releases.

PR #321, opened 2024-05-09
Move samples and packages configuration of Dokka to convention plugin

### LINKED ISSUE #210: Move Dokka setup to a separate plugin and make samples discovery flexible
Currently, Dokka configured separately in both [core](https://github.com/Kotlin/kotlinx-io/blob/develop/core/build.gradle.kts#L47) and [bytestring](https://github.com/Kotlin/kotlinx-io/blob/develop/bytestring/build.gradle.kts#L30) modules. And in each of these configs paths to files containing samples are listed explicitly.

It would be nice to:
- use file pattern instead of paths to particular files (it should be something like `**/test/samples/**/*.kt`)
- move common configuration logic to a separate plugin in `build-logic`

whyoleg:
Fixes #210

Note: I've also moved `unsafe` package hiding to convention plugin, because AFAIU the same idea of not advertising of this API in official documentation will be applied to `unsafe` APIs in `core`.


fzhinkin (approved):
@whyoleg thanks a lot for addressing the issue!

PR #327, opened 2024-05-17
Update Kotlin to 2.0

fzhinkin:


gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-kotlin = "1.9.24"
```
fzhinkin, line 2: The plan is to update it to "2.0.0" once it released, merge the branch and release kotlinx-io "0.4.0"

build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-publish.gradle.kts
```diff
@@ -92,8 +92,8 @@ fun RepositoryHandler.configureMavenPublication( project: Project) {
         }
     }
 
-    // Something that's easy to "clean" for development, not mavenLocal
-    maven("${project.rootProject.buildDir}/repo") {
+    // Something that's easy to "clean" for development not mavenLocal
+    maven("${project.layout.buildDirectory}/repo") {
```
whyoleg, line 96:
```suggestion
    maven(project.rootProject.layout.buildDirectory.dir("repo")) {
```
So that it will be published to one place in `rootProject/build` (like it was before)

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
```diff
@@ -124,6 +124,7 @@ public actual val SystemTemporaryDirectory: Path
         return Path(os.tmpdir() ?: "")
     }
 
+@Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
```
whyoleg, line 127: should be not needed after https://github.com/Kotlin/kotlinx-io/pull/325 was merged

core/native/src/files/FileSystemNative.kt
```diff
@@ -125,6 +127,7 @@ internal actual val isWindows: Boolean = Platform.osFamily == OsFamily.WINDOWS
 @OptIn(ExperimentalStdlibApi::class)
```
whyoleg, line 127: `ExperimentalStdlibApi` is not needed anymore in a lot of places where `AutoCloseable` is used (as well as `AutoCloseableAlias` is not needed anymore)
It would be nice to cleanup this before release (may be via separate PR)

core/nodeFilesystemShared/src/files/PathsNodeJs.kt
```diff
@@ -10,6 +10,7 @@ import kotlinx.io.node.buffer
 import kotlinx.io.node.fs
 import kotlinx.io.node.path as nodeJsPath
 
+@Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
```
whyoleg, line 13: same as above: should be not needed after https://github.com/Kotlin/kotlinx-io/pull/325 was merged

whyoleg (commented):
Just some minor comments :)

lppedd:
Does this change retain compatibility with 1.9.x?

core/native/src/files/FileSystemNative.kt
fzhinkin: Thanks for reminding me about that! I think it fine to clean it up here

fzhinkin:
@lppedd 
> so basically even setting
> ```
> apiVersion = KotlinVersion.KOTLIN_1_9
> languageVersion = KotlinVersion.KOTLIN_1_9
> ```
> wouldn't help?

Neither of these options affects binary compatibility, so no, unfortunately, that won't help.

>  I'm just asking that because libraries suddenly requiring 2.0+ means whoever is using 1.9.x can't update anything. I've seen coroutines doing the same, serialization too. This seems a bit risky IMO.

For kotlinx-io, the new release will only switch the Kotlin version, so those who can't update the version in their projects can continue using `0.3.5` for a while.

Note that in terms of compatibility, there's nothing new here, and the same happened previously when libraries updated the Kotlin version from 1.8 to 1.9, for example.

shanshin (approved):

PR #333, opened 2024-05-30
Add UnknownServiceException

Goooler:
I noticed Kotr is migrating to kotlinx-io in https://github.com/ktorio/ktor/pull/4032, and I sent a PR https://github.com/ktorio/ktor/pull/3725 which is depending on `UnknownServiceException`, it would be awesome to port this exception for all platforms to get rid of custom implementations in Ktor.

### DIFF
--- a/core/common/src/-CommonPlatform.kt
+++ b/core/common/src/-CommonPlatform.kt
@@ -39,6 +39,11 @@ public expect open class EOFException : IOException {
     public constructor(message: String?)
 }
 
+public expect open class UnknownServiceException : IOException {
+    public constructor()
+    public constructor(message: String?)
+}
+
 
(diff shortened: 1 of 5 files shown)


fzhinkin:
Hey @Goooler!

Thanks for opening the PR. Unfortunately, I don't think that this exception is a good fit for the kotlinx-io right now for a few reasons:
- currently, kotlinx-io provides no networking-related functionality;
- having only `UnknownServiceException` and not having other exceptions covering various networking and URN/URL-related problems _feels_ a bit off.

So for now, I'd suggest leaving the exception in the Ktor codebase.

However, in the long run, it would make sense to supply a separate package stuffed with networking-related exceptions to serve as a common types for the ecosystem.

Goooler:
Makes sense!

PR #334, opened 2024-05-30
[PR 2/5] Introduce unsafe API for bulk read/write ops

fzhinkin:
This is a first portion of unsafe API described in #135. 
This particular part aimed to facilitate integration with other frameworks and libraries.

Implemented API was initially described in the "Bulk API" subsection of https://github.com/Kotlin/kotlinx-io/issues/135#issuecomment-2125014081


fzhinkin:
CC @e5l

fzhinkin:
Previously, newly introduced API was published as a snapshot build from https://github.com/Kotlin/kotlinx-io/tree/snapshot/unsafe-api branch.
Since then, some functions changed their name, and vectorized read on JVM changed its signature slightly, but otherwise everything remained the same.

fzhinkin:
This change sits on top of https://github.com/Kotlin/kotlinx-io/pull/332.
I ran benchmark for this and develop branches to check if this PR affects existing API and current branch performs significantly better, compared to develop. That looks questionable, I'll revisit results later.

[kotlinx-io-benchmarking-result-2024-05-29-dev.json](https://github.com/Kotlin/kotlinx-io/files/15497589/kotlinx-io-benchmarking-result-2024-05-29-dev.json)
[kotlinx-io-benchmarking-result-2024-05-29-bulk-api-part-1.json](https://github.com/Kotlin/kotlinx-io/files/15497590/kotlinx-io-benchmarking-result-2024-05-29-bulk-api-part-1.json)

core/common/src/Buffer.kt
```diff
@@ -39,17 +39,26 @@ import kotlin.jvm.JvmField
  * does not affect buffer's state and [exhausted] only indicates that a buffer is empty.
  */
 public class Buffer : Source, Sink {
-    @JvmField
+    @PublishedApi
+    @get:JvmName("#int#getHead")
```
whyoleg, line 43: just curious:
1. why `JvmName` is needed here (and in all other places)
2. why it has such name pattern?

fzhinkin: To prevent use of these functions/properties from Java. Names were chosen with that in mind, `int` stands for `internal`.

All recently annotated API was, of course, directly accessible from Java before.
But as `Segment` is now public, it might feels like an invitation to use its API.

lppedd: @fzhinkin just curious, why not use `@JvmSynthetic`?

core/common/src/Segment.kt
```diff
@@ -36,46 +37,52 @@ import kotlin.jvm.JvmField
  * `limit` and beyond. There is a single owning segment for each byte array. Positions,
  * limits, prev, and next references are not shared.
  */
-internal class Segment {
+public class Segment {
```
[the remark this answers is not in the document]
fzhinkin: An API to iterate over Buffer's segments (see https://github.com/Kotlin/kotlinx-io/issues/135#issuecomment-2125014081) will be implemented in subsequent PRs. That API would expose Segments to make iteration allocation free.

core/common/src/Buffer.kt
fzhinkin: @lppedd No particular reason. Switched to `JvmSynthetic` as it seems to be a less controversial option :)

core/common/src/Buffer.kt
```diff
@@ -39,17 +39,26 @@ import kotlin.jvm.JvmField
  * does not affect buffer's state and [exhausted] only indicates that a buffer is empty.
  */
 public class Buffer : Source, Sink {
-    @JvmField
+    @PublishedApi
```
whyoleg, line 42: just curious[2]: does it make any difference (performance/code size/anything else wise) annotating it with `JvmField`? :)
so something like this:
```
@JvmField
@JvmSynthetic
@PublishedApi
internal var head: Segment? = null
```
I mean, looks like this will still work as intended.

lppedd: I've run some benchmarks on a diff library and had to push [this](https://github.com/petertrr/kotlin-multiplatform-diff/pull/112/commits/69e32f65886a463e3ea91e9e595c131cba4bfbc3) commit to solve a pretty obvious performance issue. If the field is accessed frequently I think annotating with `@JvmField` is always the best option.

fzhinkin: Having a publicly exposed field ties our hands when it comes to the future evolution of segments (or any other API). 

Any decent runtime/compiler would inline trivial methods (getters and setters usually match that criteria), so having a pair of a getter & a setter is almost the same as having a public field from a performance point of view.

core/common/src/unsafe/UnsafeBufferOperations.kt
```diff
@@ -0,0 +1,151 @@
...
+ * Copyright 2010-2024 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.io.unsafe
+
+import kotlinx.io.*
+
+@UnsafeIoApi
+public object UnsafeBufferOperations {
```
lppedd, line 11: Does it really make sense to wrap these functions in an object?
Why not going top-level?

I understand it acts as a sort-of namespace, however it means it's an additional instance, and in K/JS each function call well be behind an `undefined` check as everything is done lazily.

fzhinkin: We're trying hard to ensure users understand that the API is unsafe and they REALLY need to use it.

The necessity to call it via `UnsafeBufferOperations` is yet another hint at the call site that the API is unsafe.

>  in K/JS each function call will be behind an undefined check as everything is done lazily.

Do you expect it to be a performance bottleneck for this particular API? 
(The following is a speculation, mostly) `undefined` check should be a single comparison followed by the jump, right?

lppedd: > We're trying hard to ensure users understand that the API is unsafe and they REALLY need to use it.

Reasonable! But wouldn't the unsafe opt-in alone on each function be enough?

> Do you expect it to be a performance bottleneck for this particular API?

If we take kotlinx-io in isolation, definitely not.  
What I worry about is the build-up of lazy initialization checks. Some are contributed by io, some by coroutines, some by serialization, some by other third party libraries, etc.

That's why I always ask if it's really necessary. I hope [namespacing in the language](https://youtrack.jetbrains.com/issue/KT-11968) will remove the necessity of using objects for that.

fzhinkin: > But wouldn't the unsafe opt-in alone on each function be enough?

One can opt it in for a file or the whole project. 😞

core/common/src/Buffer.kt
lppedd: But it's the same if a user uses static-like imports:

```kotlin
object MyObject {
  fun one() {}
  fun two() {}
}
```

```kotlin
import com.test.MyObject.one
import com.test.MyObject.two

fun example() {
  one()
  two()
}
```

So I wouldn't worry too much about it.

core/common/src/unsafe/UnsafeBufferOperations.kt
lppedd: But it's the same if a user uses static-like imports:

```kotlin
object MyObject {
  fun one() {}
  fun two() {}
}
```

```kotlin
import com.test.MyObject.one
import com.test.MyObject.two

fun example() {
  one()
  two()
}
```

So I wouldn't worry too much about it.

fzhinkin: > But it's the same if a user uses static-like imports

Sure, but users have to add a project/file-wide opt-in and write a static import. That lowers the chances of accidentally using unsafe API.

It's also worth mentioning that there's no goal of making that API convenient for everyday use. It's quite the opposite: regular kotlinx-io users are not expected to use unsafe API, and a clumsier API shape is one of the leverages to nudge them toward using other APIs.

lppedd: I'll have to see the code once it's all done, but the thing you have to keep in mind is code like this

```kotlin
while (condition) {
  UnsafeBufferOperations.someFun(...)
}
```

where there are thousand+ iterations possible. That's where the `undefined` check becomes a problem.

fzhinkin: https://github.com/fzhinkin/kjs-obj-recvr-overhead/blob/main/src/commonMain/kotlin/FunctionCallBenchmark.kt

https://jmh.morethan.io/?source=https://raw.githubusercontent.com/fzhinkin/kjs-obj-recvr-overhead/main/results/2024-06-10T13.32.24.647154309-kn-js.json

It is hard to argue about an `undefined` check's impact on the performance given the high jitter, but it seems like it does not exceed the measurement error.

lppedd: For others, see [this Slack post](https://kotlinlang.slack.com/archives/CKCAAV8VB/p1718034129228599).

core/jvm/src/unsafe/UnsafeBufferOperationsJvm.kt
```diff
@@ -0,0 +1,158 @@
...
+ * @throws IllegalArgumentException when the [iovec] is empty.
+ *
+ * @sample kotlinx.io.samples.unsafe.UnsafeReadWriteSamplesJvm.gatheringWrite
+ *
+ */
+@UnsafeIoApi
+public inline fun UnsafeBufferOperations.readBulk(
+    buffer: Buffer,
+    iovec: Array<ByteBuffer?>,
+    readAction: (Array<ByteBuffer?>, Int) -> Long
```
shanshin, line 131: Same about named parameters and ULong

core/jvm/src/unsafe/UnsafeBufferOperationsJvm.kt
```diff
@@ -0,0 +1,158 @@
...
+) {
+    writeToTail(buffer, minimumCapacity) { rawData, pos, limit ->
+        val bb = ByteBuffer.wrap(rawData, pos, limit - pos).slice()
+        writeAction(bb)
+        bb.position()
+    }
+}
+
+/**
+ * Provides read-only access to [buffer]'s data by filling provided [iovec] array with [ByteBuffer]'s representing
```
shanshin, line 91: It is not entirely clear from the documentation what the size of each ByteBuffer in the array will be and, in general, on the basis of which the total number of buffers is determined, is it predictable or not, which array size is better to use

core/jvm/test/unsafe/UnsafeBufferOperationsJvmReadBulkTest.kt
```diff
@@ -0,0 +1,185 @@
...
+        }
+        val buffers = Array<ByteBuffer?>(16) { null }
+        UnsafeBufferOperations.readBulk(buffer, buffers) { array, iovecLen ->
+            assertSame(buffers, array)
+            assertEquals(4, iovecLen)
+
+            assertEquals(Segment.SIZE, array[0]!!.remaining())
+            assertEquals(Segment.SIZE, array[1]!!.remaining())
+            assertEquals(Segment.SIZE, array[2]!!.remaining())
+            assertEquals(1, array[3]!!.remaining())
```
shanshin, line 95: No content check?

core/common/src/Segment.kt
```diff
@@ -196,15 +203,55 @@ internal class Segment {
...
+     * [Segment.pos] and ends at [Segment.limit], writable slice starts at [Segment.limit]
+     * and spans over [Segment.remainingCapacity] bytes.
+     *
+     * This method exists only to preserve binary compatibility if a segment's internal
+     * container eventually changes from ByteArray to something else.
+     */
+    @PublishedApi
+    @JvmSynthetic
+    @Suppress("UNUSED_PARAMETER")
+    internal fun dataAsByteArray(readOnly: Boolean): ByteArray = data
```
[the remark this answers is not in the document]
fzhinkin: Yes, exactly

core/common/src/unsafe/UnsafeBufferOperations.kt
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Maximum value that is safe to pass to [writeToTail].
+     */
+    public const val maxSafeWriteCapacity: Int = Segment.SIZE
```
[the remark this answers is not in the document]
fzhinkin: Thanks for pointing to it. In fact, that value may depend on some settings (for example, if we support different SegmentPool configurations that could be set via env variables or Java system props). I was thinking about that case all the way, but it didn't stop me from making the property `const` 🫠

core/common/src/Segment.kt
```diff
@@ -295,3 +341,6 @@ internal fun Segment.indexOfBytesOutbound(bytes: ByteArray, startOffset: Int): I
     }
     return -1
 }
+
+@PublishedApi
+internal fun Segment.isEmpty(): Boolean = size == 0
```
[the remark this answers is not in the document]
fzhinkin: While the `size` is a core property of a `Segment` that could not be easily implemented outside the `Segment`, `isEmpty` is just a convince function. So I decided to implement it as an extension.

core/jvm/src/unsafe/UnsafeBufferOperationsJvm.kt
fzhinkin: The documentation is indeed require some improvements.

However, the API itself was designed with the idea that `iovec`-array will be allocated once and cached by the caller (for instance, as a thread-local variable) to reduce excessive allocations on the write path.
The number of ByteBuffers to be created corresponds to the number of segments a buffer has. Subsequent patches provides an API allowing to find that value, but in general, it's hard to give an advice on the array size (for example, a buffer holding 8000 bytes may have either 1 segment, or multiple).

core/common/test/unsafe/UnsafeBufferOperationsWriteTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        val buffer = Buffer()
+
+        assertFailsWith<TestException> {
+            UnsafeBufferOperations.writeToTail(buffer, 2) { _, _, _ ->
+                throw TestException()
+            }
+        }
+
+        assertTrue(buffer.exhausted())
+    }
```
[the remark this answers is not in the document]
fzhinkin: Done, thanks!

core/common/test/unsafe/UnsafeBufferOperationsWriteTest.kt
```diff
@@ -0,0 +1,131 @@
...
+    private class TestException : RuntimeException()
+
+    @Test
+    fun bufferCapacity() {
+        val buffer = Buffer()
+
+        UnsafeBufferOperations.writeToTail(buffer, 1) { data, startIndex, endIndex ->
+            // Unsafe check, head is not committed yet
+            assertSame(buffer.head!!.data, data)
+            assertEquals(buffer.head!!.data.size, (endIndex - startIndex))
```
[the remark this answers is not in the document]
fzhinkin: Removed extra checks, thanks!

core/common/test/unsafe/UnsafeBufferOperationsReadTest.kt
```diff
@@ -0,0 +1,119 @@
...
+        buffer.skip(8190)
+        val head = buffer.head!!
+        assertEquals(8190, head.pos)
+
+        UnsafeBufferOperations.readFromHead(buffer) { _, startIndex, endIndex ->
+            assertEquals(2, endIndex - startIndex)
+            2
+        }
+
+        assertEquals(9000 - 8192, buffer.size.toInt())
```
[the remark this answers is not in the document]
fzhinkin: Made them less wizardry ;)

core/jvm/test/unsafe/UnsafeBufferOperationsJvmReadFromHeadTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        val buffer = Buffer()
+        assertFailsWith<IllegalArgumentException> {
+            UnsafeBufferOperations.readFromHead(buffer) { _ -> }
+        }
+    }
+
+    @Test
+    fun readFromTheSegmentEnd() {
+        val buffer = Buffer().apply { write(ByteArray(9000) { 0xff.toByte() }) }
+        buffer.skip(8190)
```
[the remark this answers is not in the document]
fzhinkin: Fixed

core/common/test/unsafe/UnsafeBufferOperationsReadTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    private class TestException : RuntimeException()
+
+    @Test
+    fun bufferCapacity() {
+        val buffer = Buffer().apply { writeString("hello world") }
+
+        val head = buffer.head!!
+        UnsafeBufferOperations.readFromHead(buffer) { data, startIndex, endIndex ->
+            assertTrue(endIndex <= data.size)
+            assertEquals(head.size, endIndex - startIndex)
```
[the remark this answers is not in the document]
fzhinkin: Excessive checks were removed, thanks!

core/common/src/unsafe/UnsafeBufferOperations.kt
```diff
@@ -0,0 +1,151 @@
...
+     * the best effort basis, meaning that there are no strong zero-copy guarantees
+     * and the copy will be created if it could not be omitted.
+     *
+     * @throws IllegalStateException when [readAction] returns negative value or a values exceeding
+     * the `endIndex - startIndex` value.
+     * @throws IllegalArgumentException when the [buffer] is empty.
+     *
+     * @sample kotlinx.io.samples.unsafe.UnsafeBufferOperationsSamples.readByteArrayFromHead
+     */
+    public inline fun readFromHead(buffer: Buffer, readAction: (ByteArray, Int, Int) -> Int) {
```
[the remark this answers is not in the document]
fzhinkin: Done, here and in a few other places

core/jvm/test/unsafe/UnsafeBufferOperationsJvmReadBulkTest.kt
fzhinkin: Added, thanks!

core/common/src/unsafe/UnsafeBufferOperations.kt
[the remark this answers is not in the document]
fzhinkin: There are definitely no strong arguments against using `UInt` here. However, initially, I abstained from doing it for the following reasons:
- if a user consumed all the data, returned value should be calculated as a difference between two `Int` indices, and then it has to be explicitly converted to unsigned int, which is not very convenient;
- another case I found myself using quite frequently with this API is writing n-bytes into the array and then returning the last index value; it also has to be converted explicitly and it's also not very convenient;
- unsigned types adds some small overhead;
- if a user mistakengly got negative number of bytes to consume, reporting it as it is during validating might be more informative compared to some large positive unsigned value.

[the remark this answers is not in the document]
fzhinkin: So, according to https://github.com/Kotlin/KEEP/blob/master/proposals/unsigned-types.md#non-negative-integers, `UInt` does not suite well for this use case.

core/common/src/unsafe/UnsafeBufferOperations.kt
```diff
@@ -0,0 +1,151 @@
...
+     * on the best-effort basis, meaning that there are no strong zero-copy guarantees
+     * and the copy will be created if it could not be omitted.
+     *
+     * @throws IllegalStateException when [minimumCapacity] is too large and could not be fulfilled.
+     * @throws IllegalStateException when [writeAction] returns a negative value or a value exceeding
+     * the `endIndex - startIndex` value.
+     *
+     * @sample kotlinx.io.samples.unsafe.UnsafeBufferOperationsSamples.writeByteArrayToTail
+     */
+    public inline fun writeToTail(buffer: Buffer, minimumCapacity: Int, writeAction: (ByteArray, Int, Int) -> Int) {
```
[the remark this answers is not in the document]
fzhinkin: See https://github.com/Kotlin/kotlinx-io/pull/334#discussion_r1652561311

core/jvm/src/unsafe/UnsafeBufferOperationsJvm.kt
fzhinkin: See https://github.com/Kotlin/kotlinx-io/pull/334#discussion_r1652561311

shanshin (approved):

PR #339, opened 2024-06-10
[PR 5/5] Hide Segment::data

fzhinkin:
This PR concludes unsafe-API-related changes by hiding the `Segment::data.`

The change itself has nothing to do with unsafe API as it is now but ensures that direct internal data won't leak into the codebase.


core/common/src/Buffer.kt
```diff
@@ -624,20 +571,20 @@ public class Buffer : Source, Sink {
...
+        UnsafeBufferOperations.iterate(this) { ctx, head ->
+            var bytesWritten = 0
+            var seg: Segment? = head
+            do {
+                seg!!
+                var idx = 0
+                while (bytesWritten < len && idx < seg.size) {
+                    val b = ctx.getUnchecked(seg, idx++)
+                    bytesWritten++
+                    builder.append(HEX_DIGIT_CHARS[(b shr 4) and 0xf])
```
shanshin, line 583: Same here, append byte-by-byte

core/common/src/Buffer.kt
```diff
@@ -80,119 +81,100 @@ public class Buffer : Source, Sink {
...
+        }
+        val segmentSize = segment.size
+        if (segmentSize < 8) {
+            // If the short is split across multiple segments, delegate to readInt().
+            require(8)
+            if (segmentSize == 0) {
+                recycleHead()
+                return readLong()
+            }
+            return (readInt().toLong() shl 32 or (readInt().toLong() and 0xffffffffL))
```
shanshin, line 164: Is this way of reading exactly faster than using Unsafe API?

It turns out that in the worst case, we will fall down the call stack by an additional 3 frames down

core/common/src/Segment.kt
```diff
@@ -185,6 +184,84 @@ public class Segment {
...
+        return i
+    }
+
+    internal fun readLong(): Long {
+        val data = data
+        var pos = pos
+        val v = (
+                data[pos++] and 0xffL shl 56
+                        or (data[pos++] and 0xffL shl 48)
+                        or (data[pos++] and 0xffL shl 40)
```
shanshin, line 254: Maybe it's paranoia, but does it work at the same speed as 
`or (data[pos + 1] and 0xffL shl 40)`
?

(there is no frequent updating of the field value)

fzhinkin: `pos` here is a local variable defined few lines above, so the field will be updated only once, after all reads are done.

Anyway, I'll re-check how it works with explicit offsets and immutable base, the last time I tested it, there was no significant difference.

core/common/src/Buffer.kt
fzhinkin: `Buffer.toString` is not expected to be used for anything but debugging (as it can only capture a prefix of a buffer)

core/common/src/Buffer.kt
fzhinkin: Frankly, I didn't optimize the cross-segment read/write scenarios as they should not occur as frequently as reads/writes fitting inside a single segment.

As the current implementation should not make it worse, I suggest investigating and optimizing (if needed) these sites separately.

core/common/src/Buffer.kt
```diff
@@ -80,119 +81,100 @@ public class Buffer : Source, Sink {
...
+        if (segmentSize == 8) {
             recycleHead()
-        } else {
-            segment.pos = pos
         }
-
         return v
     }
 
+    private fun throwEof(byteCount: Long) {
```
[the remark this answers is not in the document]
fzhinkin: Thanks for suggestion, works like a charm!

shanshin (approved):

PR #352, opened 2024-07-04
SegmentPool improvements

fzhinkin:
Currently, SegmentPool implementation on JVM works more like a cache than a pool: it has a relatively small fixed size, segments are taken from the pool on an effort basis, and it's easy to make a segment non-recyclable (meaning that it will never be returned to the pool).

This behavior is totally fine for many applications, but it's not for Ktor.
This PR addresses several issues:
- Support precise segment reference tracking to return shared segments back to the pool eventually. Currently, once a segment is shared, it is never returned to the pool because the segment state is tracked by a flag that could be set but could not be used. To solve that problem, on JVM, the exact number of shared copies is now tracked, and the value is decremented on attempts to recycle shared copies. Once the last copy is passed to the `Segment.recycle`, it'll be returned to the pool.
- Retry lost CAS attempts when taking or returning segments from the pool. The previous sentence is quite descriptive; I'll only add that it does not make much sense without the next change.
- Support segment pool size configuration. Currently, the segment pool consists of multiple buckets (the exact number depends on the system's logical CPU count), each with a max size of 64KiB. This PR adds a second level/tier to the pool with half as many buckets as an old first tier. The total size of the new tier is configurable via system properties and is `0` by default. All failed attempts to take or recycle segments in the first tier use the second tier as a fallback. Unlike the first tier, if the second tier's bucket is empty, `take` will scan all other buckets. The scan is also performed on `recycle` to a full bucket.

All these changes reduce the allocation count by combining more precise work with the pool and optionally extensible pool size.
Two-level design helps with avoiding performance penalties induced by the scan technique employed by the second tier when that tier is not used and, at the same time, allows using the same sharding technique as in the first tier enhanced with the bucket steeling techniques.

### DIFF
--- a/core/common/src/SegmentPool.kt
+++ b/core/common/src/SegmentPool.kt
@@ -38,4 +38,6 @@ internal expect object SegmentPool {
 
     /** Recycle a segment that the caller no longer needs. */
     fun recycle(segment: Segment)
+
+    fun tracker(): SegmentCopyTracker
 }
--- a/core/common/src/SegmentPool.kt
+++ b/core/common/src/SegmentPool.kt
@@ -39,5 +39,10 @@ internal expect object SegmentPool {
     /** Recycle a segment that the caller no longer needs. */
     fun recycle(segment: Segment)
 
+    /**
+     * Allocates a new copy tracker that'll be associated with a segment.
+     * For performance reasons, there's no tracker attached to a segment initially.
+     * Instead, it's allocated lazily on the first sharing attempt.
+     */
     fun tracker(): SegmentCopyTracker
 }
--- a/core/common/src/SegmentPool.kt
+++ b/core/common/src/SegmentPool.kt
@@ -40,7 +40,7 @@ internal expect object SegmentPool {
     fun recycle(segment: Segment)
 
     /**
-     * Allocates a new copy tracker that'll be associated with a segment.
+     * Allocates a new copy tracker that'll be associated with a segment from this pool.
      * For performance reasons, there's no tracker attached to a segment initially.
      * Instead, it's allocated lazily on the first sharing attempt.
      */
--- a/core/common/src/unsafe/UnsafeBufferOperations.kt
+++ b/core/common/src/unsafe/UnsafeBufferOperations.kt
@@ -38,7 +38,7 @@ public object UnsafeBufferOperations {
         checkBounds(bytes.size, startIndex, endIndex)
         val segment = Segment.new(
             bytes, startIndex, endIndex,
-            SimpleCopyTracker().also { it.addCopy() } /* to prevent recycling */,
+            AlwaysSharedCopyTracker, /* to prevent recycling */
             owner = false /* can't append to it */
         )
         val tail = buffer.tail
--- a/core/common/src/unsafe/UnsafeBufferOperations.kt
+++ b/core/common/src/unsafe/UnsafeBufferOperations.kt
@@ -37,7 +37,8 @@ public object UnsafeBufferOperations {
     public fun moveToTail(buffer: Buffer, bytes: ByteArray, startIndex: Int = 0, endIndex: Int = bytes.size) {
         checkBounds(bytes.size, startIndex, endIndex)
         val segment = Segment.new(
-            bytes, startIndex, endIndex, shared = true /* to prevent recycling */,
+            bytes, startIndex, endIndex,
+            SimpleCopyTracker().also { it.addCopy() } /* to prevent recycling */,
             owner = false /* can't append to it */
         )
         val tail = buffer.tail
--- a/core/jvm/src/SegmentPool.kt
+++ b/core/jvm/src/SegmentPool.kt
@@ -126,13 +126,13 @@ internal actual object SegmentPool {
         when {
             first === LOCK -> {
                 // We didn't acquire the lock. Don't take a pooled segment.
-                return Segment.new(RefCountingCopyTracker())
+                return Segment.new()
             }
 
             first == null -> {
                 // We acquired the lock but the pool was empty. Unlock and return a new segment.
                 firstRef.set(null)
-                return Segment.new(RefCountingCopyTracker())
+                return Segment.new()
             }
 
             else -> {
@@ -149,7 +149,7 @@ internal actual object SegmentPool {
     @JvmStatic
     actual fun recycle(segment: Segment) {
         require(segment.next == null && segment.prev == null)
-        if (segment.copyTracker.removeCopyIfShared()) return // This segment cannot be recycled.
+        if (segment.copyTracker?.removeCopyIfShared() == true) return // This segment cannot be recycled.
 
         val firstRef = firstRef()
 
@@ -168,6 +168,9 @@ internal actual object SegmentPool {
             segment.next = null // Don't leak a reference in the pool either!
         }
     }
+
+    @JvmStatic
+    actual fun tracker(): SegmentCopyTracker = RefCountingCopyTracker()
 
     private fun firstRef(): AtomicReference<Segment?> {
         // Get a value in [0..HASH_BUCKET_COUNT) based on the current thread.
--- a/core/jvm/src/SegmentPool.kt
+++ b/core/jvm/src/SegmentPool.kt
@@ -67,24 +67,28 @@ internal class RefCountingCopyTracker : SegmentCopyTracker() {
 
 /**
  * This class pools segments in a lock-free singly-linked stack. Though this code is lock-free it
- * does use a sentinel [LOCK] value to defend against races. Conflicted operations are not retried,
- * so there is no chance of blocking despite the term "lock".
+ * does use a sentinel [LOCK] value to defend against races. To reduce the contention, the pool consists
+ * of several buckets (see [HASH_BUCKET_COUNT]), each holding a reference to its own segments stack.
+ * Every [take] or [recycle] choose one of the buckets depending on a [Thread.currentThread]'s [Thread.getId].
  *
  * On [take], a caller swaps the stack's next pointer with the [LOCK] sentinel. If the stack was
  * not already locked, the caller replaces the head node with its successor.
  *
  * On [recycle], a caller swaps the head with a new node whose successor is the replaced head.
  *
- * On conflict, operations succeed, but segments are not pushed into the stack. For example, a
- * [take] that loses a race allocates a new segment regardless of the pool size. A [recycle] call
- * that loses a race will not increase the size of the pool. Under significant contention, this pool
- * will have fewer hits and the VM will do more GC and zero filling of arrays.
+ * On conflict, operations are retried until they succeed.
  *
  * This tracks the number of bytes in each linked list in its [Segment.limit] property. Each element
  * has a limit that's one segment size greater than its successor element. The maximum size of the
  * pool is a product of [MAX_SIZE] and [HASH_BUCKET_COUNT].
  *
- * TODO: update kdoc with info about L2 pool
+ * [MAX_SIZE] is kept relatively small to avoid excessive memory consumption in case of a large [HASH_BUCKET_COUNT].
+ * For better handling of scenarios with high segments demand, an optional second-level pool could be enabled
+ * by setting up a value of `kotlinx.io.pool.size.bytes` system property.
+ *
+ * The second-level pool, unlike the pool described above, is not sharded and has a single entry point shared
+ * across all application threads. That pool is used as a backup in case when [take] or [recycle] failed due to
+ * an empty or exhausted segments chain in a corresponding bucket (one of [HASH_BUCKET_COUNT] buckets).
  */
 internal actual object SegmentPool {
     /** The maximum number of bytes to pool per hash bucket. */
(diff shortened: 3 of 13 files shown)


fzhinkin:
CC @e5l @bjhham

core/jvm/src/SegmentPool.kt
```diff
@@ -62,6 +109,14 @@ internal actual object SegmentPool {
     private val HASH_BUCKET_COUNT =
         Integer.highestOneBit(Runtime.getRuntime().availableProcessors() * 2 - 1)
 
+    private val HASH_BUCKET_COUNT_L2 = (HASH_BUCKET_COUNT / 2).coerceAtLeast(1)
+
+    private val SECOND_LEVEL_POOL_TOTAL_SIZE =
+        System.getProperty("kotlinx.io.pool.size.bytes", "0").toInt().coerceAtLeast(0)
```
fzhinkin, line 115: Discussed with @e5l that it makes sense to bump up the default size up to several megs (probably, 4Mb).

Not sure if it should be done on Android.

core/jvm/src/SegmentPool.kt
```diff
@@ -25,34 +25,81 @@ import kotlinx.io.SegmentPool.LOCK
...
  */
 internal actual object SegmentPool {
     /** The maximum number of bytes to pool per hash bucket. */
     // TODO: Is 64 KiB a good maximum size? Do we ever have that many idle segments?
     actual val MAX_SIZE = 64 * 1024 // 64 KiB.
 
     /** A sentinel segment to indicate that the linked list is currently being modified. */
-    private val LOCK = Segment.new(ByteArray(0), pos = 0, limit = 0, shared = false, owner = false)
+    private val LOCK =
+        Segment.new(ByteArray(0), pos = 0, limit = 0, copyTracker = SimpleCopyTracker(), owner = false)
```
qwwdfsad, line 102: tracker can be null here?

core/jvm/src/SegmentPool.kt
```diff
@@ -73,6 +128,10 @@ internal actual object SegmentPool {
         AtomicReference<Segment?>() // null value implies an empty bucket
     }
 
+    private val hashBucketsL2: Array<AtomicReference<Segment?>> = Array(HASH_BUCKET_COUNT) {
```
qwwdfsad, line 131: Consider using `AtomicReferenceArray` instead

core/common/src/SegmentPool.kt
```diff
@@ -38,4 +38,6 @@ internal expect object SegmentPool {
 
     /** Recycle a segment that the caller no longer needs. */
     fun recycle(segment: Segment)
+
+    fun tracker(): SegmentCopyTracker
```
qwwdfsad, line 42: Consider adding a small KDoc

core/jvm/src/SegmentPool.kt
```diff
@@ -83,55 +142,148 @@ internal actual object SegmentPool {
...
 
+    @JvmStatic
+    actual fun tracker(): SegmentCopyTracker = RefCountingCopyTracker()
+
     private fun firstRef(): AtomicReference<Segment?> {
         // Get a value in [0..HASH_BUCKET_COUNT) based on the current thread.
         @Suppress("DEPRECATION") // TODO: switch to threadId after JDK19
-        val hashBucket = (Thread.currentThread().id and (HASH_BUCKET_COUNT - 1L)).toInt()
-        return hashBuckets[hashBucket]
+        val bucket = (Thread.currentThread().id and (HASH_BUCKET_COUNT - 1L)).toInt()
```
qwwdfsad, line 280: Filed https://youtrack.jetbrains.com/issue/KT-69804/Introduce-Thread.threadId-that-encapsulates-either-Thread.getId-or-Thread.threadId and https://youtrack.jetbrains.com/issue/KT-69803/Support-multi-release-JAR-capabilities-in-the-standard-library

qwwdfsad (commented):
Round of cosmetics to speed up the convergence, still digging into the segment pool details

core/jvm/src/SegmentPool.kt
qwwdfsad: What really puzzles me here is how it affects the end users.
Because it seems like the characteristics of the IO (which is *the* way to do a lot of stuff) will depend on whether or not Ktor is used in the project.

I don't mind it as a palliative intermediate solution as the ground for further benchmarking, but definitely not something to keep as as

core/jvm/src/SegmentPool.kt
```diff
@@ -25,34 +25,81 @@ import kotlinx.io.SegmentPool.LOCK
...
+
+    override fun addCopy() {
+        fieldUpdater.incrementAndGet(this)
+    }
+
+    override fun removeCopyIfShared(): Boolean {
+        while (true) {
+            val value = refs
+            check(value >= 0) { "Shared copies count is negative: $value" }
+            if (fieldUpdater.compareAndSet(this, value, (value - 1).coerceAtLeast(0))) {
```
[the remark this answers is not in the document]
fzhinkin: When returning a non-sharing segment into a pool, its `refs` value (if a tracker was allocated at all) will be `0` and it would be incorrect to decrement it further.

core/jvm/src/SegmentPool.kt
fzhinkin: Sure

core/jvm/src/SegmentPool.kt
```diff
@@ -83,55 +142,148 @@ internal actual object SegmentPool {
...
-            }
+        while (true) {
+            when (val first = firstRef.getAndSet(LOCK)) {
+                LOCK -> {
+                    // We didn't acquire the lock. Let's try again
+                    continue
+                }
+
+                null -> {
+                    // We acquired the lock but the pool was empty. Unlock and return a new segment.
```
qwwdfsad, line 153: This comment contradicts the flow -- we unlock the bucket *and maybe fall through* into the second-level cache. 
It is guarded by the lock on its own, but still a ground to make mistakes

core/common/src/SegmentPool.kt
fzhinkin: Done, please check

core/common/src/SegmentPool.kt
```diff
@@ -38,4 +38,11 @@ internal expect object SegmentPool {
 
     /** Recycle a segment that the caller no longer needs. */
     fun recycle(segment: Segment)
+
+    /**
+     * Allocates a new copy tracker that'll be associated with a segment.
```
qwwdfsad, line 43:
```suggestion
     * Allocates a new copy tracker that'll be associated with a segment from this pool.
```

core/jvm/src/SegmentPool.kt
fzhinkin: Fixed, thanks for pointing out

core/jvm/src/SegmentPool.kt
fzhinkin: I checked how `BufferReadWriteByteArray` performs depending on the presence of the second-level pool and the working set size.

If WSS is within L1-pool, there's no difference.
If WSS doesn't fit into L1-pool, it might be beneficial not to use L2-pool (i.e. it's faster w/o L2-pool) until some threshold size, but then L2-pool boosts the performance.
But it's hard to judge by looking at synthetic benchmarks only, as they provide zero to no evidence regarding GC-related effects on overall performance.

core/jvm/test/PoolingTest.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package kotlinx.io
+
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class PoolingTest {
+    @Test
+    fun segmentSharing() {
```
shanshin, line 13: Should we also add tests for using Byte and writing one buffer to another?
*and Buffer.copy

core/jvm/src/SegmentPool.kt
```diff
@@ -25,34 +25,81 @@ import kotlinx.io.SegmentPool.LOCK
...
+ * The class is internal for testing only.
+ */
+internal class RefCountingCopyTracker : SegmentCopyTracker() {
+    companion object {
+        @JvmStatic
+        private val fieldUpdater = AtomicIntegerFieldUpdater.newUpdater(RefCountingCopyTracker::class.java, "refs")
+    }
+
+    @Volatile
+    private var refs: Int = 0
```
shanshin, line 46: Please, rename property to `copyCount` or something like that, because by the name of `refs` it seems that this is the total number of segments

core/common/src/unsafe/UnsafeBufferOperations.kt
```diff
@@ -37,7 +37,8 @@ public object UnsafeBufferOperations {
     public fun moveToTail(buffer: Buffer, bytes: ByteArray, startIndex: Int = 0, endIndex: Int = bytes.size) {
         checkBounds(bytes.size, startIndex, endIndex)
         val segment = Segment.new(
-            bytes, startIndex, endIndex, shared = true /* to prevent recycling */,
+            bytes, startIndex, endIndex,
+            SimpleCopyTracker().also { it.addCopy() } /* to prevent recycling */,
```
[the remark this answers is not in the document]
fzhinkin: Nice catch! Indeed, that should work better

core/jvm/src/SegmentPool.kt
```diff
@@ -62,76 +109,188 @@ internal actual object SegmentPool {
...
+    private val HASH_BUCKET_COUNT_L2 = (HASH_BUCKET_COUNT / 2).coerceAtLeast(1)
+
+    // For now, keep things on Android as they were before, but on JVM - use second level cache.
+    // See https://developer.android.com/reference/java/lang/System#getProperties() for property name.
+    private val DEFAULT_SECOND_LEVEL_POOL_TOTAL_SIZE = when (System.getProperty("java.vm.name")) {
+        "Dalvik" -> "0"
+        else -> "4194304" // 4MB
+    }
+
+    private val SECOND_LEVEL_POOL_TOTAL_SIZE =
```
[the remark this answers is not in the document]
fzhinkin: > when is the first usage of class SegmentPool?

On a first write into a buffer

> Is there a high chance of programmatically changing this value before its first reading?

I think it's achievable by updating the system property before SegmentPool is initialized.

core/jvm/src/SegmentPool.kt
```diff
@@ -62,76 +109,188 @@ internal actual object SegmentPool {
...
+            if (firstLimit >= MAX_SIZE) {
+                // L1 pool is full.
+                if (SECOND_LEVEL_POOL_TOTAL_SIZE > 0) {
+                    recycleL2(segment)
+                }
+                return
+            }
+
+            segment.next = first
+            segment.limit = firstLimit + Segment.SIZE
```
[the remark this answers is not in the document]
fzhinkin: That's correct. I'll fix the check, thanks for spotting!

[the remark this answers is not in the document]
fzhinkin: But that's a problem only for the second level pool.

core/jvm/src/SegmentPool.kt
```diff
@@ -62,76 +109,188 @@ internal actual object SegmentPool {
...
+    // For now, keep things on Android as they were before, but on JVM - use second level cache.
+    // See https://developer.android.com/reference/java/lang/System#getProperties() for property name.
+    private val DEFAULT_SECOND_LEVEL_POOL_TOTAL_SIZE = when (System.getProperty("java.vm.name")) {
+        "Dalvik" -> "0"
+        else -> "4194304" // 4MB
+    }
+
+    private val SECOND_LEVEL_POOL_TOTAL_SIZE =
+        System.getProperty("kotlinx.io.pool.size.bytes", DEFAULT_SECOND_LEVEL_POOL_TOTAL_SIZE)
+            .toInt()
```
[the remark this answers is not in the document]
fzhinkin: I'll use our favorite `OrNull` function, thanks!

core/jvm/test/PoolingTest.kt
fzhinkin: There's already a test case covering the `.copy()`, if I got you right.
Added a case with `.write`.

core/jvm/src/SegmentPool.kt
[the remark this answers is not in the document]
fzhinkin: Reworked this part after Sergey's suggestions.

core/jvm/src/SegmentPool.kt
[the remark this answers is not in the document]
shanshin: > I think it's achievable by updating the system property before SegmentPool is initialized.

if this is acceptable, then fine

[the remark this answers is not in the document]
fzhinkin: I don't see any severe consequences of that. That's the only kind of _dynamic configuration_  (https://github.com/Kotlin/kotlinx-io/pull/352#discussion_r1672341359) we can provide for now. :)

shanshin (approved):

qwwdfsad (approved):

PR #364, opened 2024-08-02
Propagate byte read/write counts from unsafe buffer ops

### LINKED ISSUE #360: `readFromHead`/`writeToTail` should propagate the `Int` returned from the lambda
I'm trying to implement `RawSource.readAtMostTo` using `UnsafeBufferOperations.writeToTail`.

I want to do
```kotlin
		return UnsafeBufferOperations.writeToTail(sink, 1) { bytes, startIndexInclusive, endIndexExclusive ->
			val read: Int = TODO()
			read
		}.toLong()
	}
```
but `writeToTail` doesn't give me the `Int` which I return from the lambda at the callsite in order to do this.

JakeWharton:
This allows the caller to more easily perform their own bookkeeping based on the amount of bytes moved.

Closes #360


fzhinkin (changes requested):
The change itself looks good, but I think we should also update JVM-specific extensions.
For `readFromHead` and `writeToTail` the change should be trivial as they delegate to common functions.
For `readBulk` it should be trivial too, we just have to use `Long` instead of `Int` there.

JakeWharton:
Indeed. Done.

fzhinkin:
@e5l FYI: formally, that's a breaking change, but Ktor (the only user of the API published on GH) should be unaffected since updated functions are all inline functions.

fzhinkin (approved):
LGTM. Thanks!

PR #367, opened 2024-08-02
Add calls-in-place contracts to unsafe operations

### LINKED ISSUE #361: Lambda-accepting `UnsafeBufferOperations` need contracts for calls-in-place
Kotlin 2.0 can see through `inline fun`s now, but it does not infer calls-in-place.

One workaround for #360 would be to do
```kotlin
		val bytesRead: Int
		UnsafeBufferOperations.writeToTail(sink, 1) { bytes, startIndexInclusive, endIndexExclusive ->
			val read: Int = TODO()
			bytesRead = read
			read
		}
		return bytesRead.toLong()
	}
```
except without the contract the compiler complains `bytesRead` is not initialized at the return.

JakeWharton:
Closes #361

### DIFF
--- a/core/common/src/Buffer.kt
+++ b/core/common/src/Buffer.kt
@@ -20,6 +20,9 @@
  */
 package kotlinx.io
 
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind.EXACTLY_ONCE
+import kotlin.contracts.contract
 import kotlin.jvm.JvmSynthetic
 
 /**
@@ -706,11 +709,18 @@ public class Buffer : Source, Sink {
  */
 @PublishedApi
 @JvmSynthetic
+@OptIn(ExperimentalContracts::class)
 internal inline fun <T> Buffer.seek(
     fromIndex: Long,
     lambda: (Segment?, Long) -> T
 ): T {
-    if (this.head == null) lambda(null, -1L)
+    contract {
+        callsInPlace(lambda, EXACTLY_ONCE)
+    }
+
+    if (this.head == null) {
+        return lambda(null, -1L)
+    }
 
     if (size - fromIndex < fromIndex) {
         var s = tail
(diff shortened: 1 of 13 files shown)


core/common/src/Buffer.kt
```diff
@@ -706,11 +709,18 @@ public class Buffer : Source, Sink {
...
     lambda: (Segment?, Long) -> T
 ): T {
-    if (this.head == null) lambda(null, -1L)
+    contract {
+        callsInPlace(lambda, EXACTLY_ONCE)
+    }
+
+    if (this.head == null) {
+        return lambda(null, -1L)
+    }
```
JakeWharton, line 723: This seems like it was a bug. Prior to this, the lambda would be called twice. Once with `(null, -1)` and then again in the `else` with `(null, 0)`. Contract validation failed without changing the implementation to early-return here, as I expect was intended.

fzhinkin: It was a bug, indeed. Thanks for catching and fixing it!

core/common/src/unsafe/UnsafeBufferOperations.kt
```diff
@@ -85,6 +90,10 @@ public object UnsafeBufferOperations {
         buffer: Buffer,
         readAction: (bytes: ByteArray, startIndexInclusive: Int, endIndexExclusive: Int) -> Int
     ) {
+        contract {
```
fzhinkin, line 93: @JakeWharton could you please also plug the same contracts into extension functions declared in `UnsafeBufferOpertationsJvm.kt`?

fzhinkin (changes requested):

core/common/src/unsafe/UnsafeBufferOperations.kt
JakeWharton: Yep!

Since these surely will conflict with #364, I'll wait for that to land then rebase & update this one.

JakeWharton: Rebased and updated.

fzhinkin (approved):
LGTM. Thanks for the fix, @JakeWharton!

PR #374, opened 2024-08-16
Eliminate more trivial backing 'val' storage

JakeWharton:


fzhinkin (approved):
Nice! Thanks for spotting and fixing the issue.

MohammadKHC:
Thanks for this in-depth explanation!
I think this makes sense.
You are awesome! Thanks again!

PR #376, opened 2024-08-21
Minor cleanup and simplification of readString

qwwdfsad:
* Request Long.MAX_VALUE from the source at once (note: no need for a loop as we are unable to build the string of such size anyway)
* Get rid of redundant checks in commonReadUtf8

Ticks one of the boxes in #342

### DIFF
--- a/benchmarks/src/commonMain/kotlin/ReadStringBenchmark.kt
+++ b/benchmarks/src/commonMain/kotlin/ReadStringBenchmark.kt
@@ -17,7 +17,7 @@ import kotlin.random.Random
 @State(Scope.Benchmark)
 open class ReadStringBenchmark() {
 
-    @Param("16", "64", "512")
+    @Param("16", "64", "512") // Fits into a single segment, so the benchmark does not measure segment boundaries crossing
     var size: Int = 0
 
     val buffer: Buffer = Buffer()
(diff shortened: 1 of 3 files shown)


benchmarks/src/commonMain/kotlin/ReadStringBenchmark.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlinx.io.writeCodePointValue
+import kotlinx.io.writeString
+import kotlin.random.Random
+
+
+@State(Scope.Benchmark)
+open class ReadStringBenchmark() {
+
+    @Param("16", "64", "512")
+    var size: Int = 0
```
fzhinkin, line 21: Did you intentionally pick size values so that a string always fit into a single segment? If so, could you please drop a comment in the code.

It would be also nice to mention why we have this benchmark along with `Utf8StringBenchmark`.

fzhinkin (changes requested):
@qwwdfsad thanks for the change! Everything looks good in general, I have only a few nits regarding the benchmark.

fzhinkin (approved):
LGTM

benchmarks/src/commonMain/kotlin/ReadStringBenchmark.kt
qwwdfsad: Yes, dropped a comment.

>It would be also nice to mention why we have this benchmark along with Utf8StringBenchmark.

My general approach is the following -- there are "generic" benchmarks that are the overall proxy to the performance of a particular method/pipeline. E.g. full-blown JSON transformation, various modes of UTF decoding etc. They are good to measure the impact of the significant change.

On the other hand, there are "tailored" ones that are written to specifically pinpoint some code-pattern or behavior.

`Utf8StringBenchmark` is the former, this one is the latter. I do not mind changing it as you see fit

PR #379, opened 2024-08-27
Improve segment pool test coverage

### LINKED ISSUE #356: Improve SegmentPool test coverage
Currently, there are a few naïve tests checking how `SegmentPool` works on JVM.
Unfortunately, these tests can't catch up with increasing complexity of the underlying implementation.

fzhinkin:
Closes #356

### DIFF
--- a/core/common/src/Segment.kt
+++ b/core/common/src/Segment.kt
@@ -20,6 +20,7 @@
  */
 package kotlinx.io
 
+import kotlin.concurrent.Volatile
 import kotlin.jvm.JvmField
 import kotlin.jvm.JvmSynthetic
 
--- a/core/common/src/Segment.kt
+++ b/core/common/src/Segment.kt
@@ -20,7 +20,6 @@
  */
 package kotlinx.io
 
-import kotlin.concurrent.Volatile
 import kotlin.jvm.JvmField
 import kotlin.jvm.JvmSynthetic
 
--- a/core/jvm/test/PoolingTest.kt
+++ b/core/jvm/test/PoolingTest.kt
@@ -5,6 +5,10 @@
 
 package kotlinx.io
 
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import kotlin.random.Random
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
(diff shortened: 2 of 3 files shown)


core/common/src/Segment.kt
```diff
@@ -20,6 +20,7 @@
  */
 package kotlinx.io
 
+import kotlin.concurrent.Volatile
```
shanshin, line 23: Unused?

core/jvm/test/PoolingTest.kt
```diff
@@ -40,4 +45,107 @@ class PoolingTest {
...
+        }
+        segments2.forEach(SegmentPool::recycle)
+
+        assertEquals(segments, segments2)
+    }
+
+    @Test
+    fun contendedUseTest() {
+        val threadsCount = Runtime.getRuntime().availableProcessors()
+        val segmentsPerThread = SegmentPool.SECOND_LEVEL_POOL_TOTAL_SIZE / Segment.SIZE / threadsCount
```
shanshin, line 82: It turns out that we only take into account fully filled segments?

core/jvm/test/PoolingTest.kt
```diff
@@ -40,4 +45,107 @@ class PoolingTest {
...
+                        observedSegments[seg] = seg
+                    }
+                    segments.forEach { SegmentPool.recycle(it) }
+                }
+            }
+        }
+        pool.shutdown()
+        pool.awaitTermination(Long.MAX_VALUE, TimeUnit.SECONDS)
+
+        val maxPoolSize =
```
shanshin, line 102: Should we try to limit the number of segments more precisely?
E.g. use `(SegmentPool.SECOND_LEVEL_POOL_TOTAL_SIZE + SegmentPool.MAX_SIZE * threadsCount) / Segment.SIZE`

core/jvm/test/PoolingTest.kt
```diff
@@ -40,4 +45,107 @@ class PoolingTest {
...
+            (SegmentPool.SECOND_LEVEL_POOL_TOTAL_SIZE + SegmentPool.MAX_SIZE * SegmentPool.HASH_BUCKET_COUNT) / Segment.SIZE
```
shanshin, line 148: same as above

core/jvm/test/PoolingTest.kt
fzhinkin: Pool size reflects maximum number of segments it'll hold once they returned back into pool. When segments returned, they are all empty, so their whole size contributes into the pool size.

core/jvm/test/PoolingTest.kt
fzhinkin: Instead, I increased the number of threads to make sure that it'll exceed the number of buckets

shanshin (approved):

PR #386, opened 2024-09-05
Convert ByteString from/to Int8Array

### LINKED ISSUE #268: Add `ByteString` conversions for JS binary types like `ArrayBuffer`, `Int8Array`, `Blob`, etc.
In the same vibe as https://github.com/Kotlin/kotlinx-io/issues/266, we could add conversions from JS binary types like `ArrayBuffer`, `Blob`, `Int8Array`...

Dealing with binary data in JS is a bit annoying. For example, the `ArrayBuffer` type is not directly usable and needs to be wrapped in a view. Providing such conversions out of the box in the `kotlinx-io-bytestring` artifact would be useful for a true multiplatform experience.

Examples:

```kotlin
fun ByteString.toArrayBuffer(): ArrayBuffer = toInt8Array().buffer

fun ByteString.toInt8Array() = Int8Array(toArrayOfBytes())

private fun ByteString.toArrayOfBytes() = Array(size) { this[it] }

fun ArrayBuffer.toByteString(): ByteString = ByteString.wrap(toByteArray())

fun ArrayBuffer.toByteArray(): ByteArray = Int8Array(this).toByteArray()

fun Int8Array.toByteArray() = ByteArray(length) { this[it] }
```

fzhinkin:
Closes #268

### DIFF
--- a/bytestring/js/src/ByteStringsJs.kt
+++ b/bytestring/js/src/ByteStringsJs.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. and respective authors and developers.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENCE file.
+ */
+
+package kotlinx.io.bytestring
+
+import org.khronos.webgl.ArrayBuffer
+import org.khronos.webgl.Int8Array
+
+/**
+ * Returns a new [Int8Array] instance initialized with bytes copied from [this] ByteString.
+ */
+public fun ByteString.toInt8Array(): Int8Array = toByteArray().unsafeCast<Int8Array>()
+
+/**
+ * Returns a new [ArrayBuffer] instance initialized with bytes copied from [this] ByteString.
+ */
+public fun ByteString.toArrayBuffer(): ArrayBuffer = toInt8Array().buffer
+
+/**
+ * Return a new [ByteString] holding data copied from [this] Int8Array.
+ */
+public fun Int8Array.toByteString(): ByteString = ByteString(*unsafeCast<ByteArray>())
+
+/**
+ * Returns a new [ByteString] holding data copied from [this] ArrayBuffer
+ */
+public fun ArrayBuffer.toByteString(): ByteString = Int8Array(this).toByteString()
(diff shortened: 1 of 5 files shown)


bytestring/js/src/ByteStringsJs.kt
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. and respective authors and developers.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENCE file.
+ */
+
+package kotlinx.io.bytestring
+
+import org.khronos.webgl.ArrayBuffer
```
lppedd, line 8: This is something that should probably be addressed for all kotlinx libraries: should everyone use these built-in types, which are pretty weird considering we generally don't want anything to do with WebGL, or should they come from kotlin-wrappers, which tend to be the most used?

The question is important because the only "interop" between same types coming from different libs is unsafe-casting.

fzhinkin: If an `ArrayBuffer` originating from `kotlin-wrappers` needs to be converted to/from `ByteString`, then it should be done either in `kotlin-wrappers`, or in a separate library.

lppedd: It's reasonable. However I'm thinking that since most people on K/JS are using kotlin-wrappers (and most commonly its `kotlin-js` module), it would be a surprise to them that the code is red in the IDE when calling `Int8Array.toByteString()`, for example.

It's a bit of an unfortunate situation because it's all fragmented into multiple places.

1. The `org.khronos.webgl` package is explicitly [documented](https://github.com/JetBrains/kotlin/blob/5f6fd3b748204bd5b054310acb56cafd3de6fbf4/libraries/stdlib/src/Module.md?plain=1#L162) as a wrapper for WebGL, so the question "why am I using WebGL types?" would come up fairly naturally.

2. The `org.kronos.webgl` package is a remnant of the extraction ([KT-35973](https://youtrack.jetbrains.com/issue/KT-35973)) of DOM APIs from the stdlib, and there is also an open issue about moving away from it ([KT-57303](https://youtrack.jetbrains.com/issue/KT-57303)). So maybe using kotlin-wrapper types ensures better stability on the long run.

lppedd: > since most people on K/JS are using kotlin-wrappers

After using kotlinx-wrappers for a while, and after getting rid of them and hosting my own types for compatibility reasons, I think another topic to consider is "multiplatformability".  
By that I mean how easy it is to have a common, let's say, `ByteArray` that actualizes to an `Int8Array` under JS, and to other types in other platforms.

kotlin-wrappers typed arrays are now [reflecting](https://github.com/JetBrains/kotlin-wrappers/blob/96043a7f734fda1c5e5b8de0ed766847f8fdccc8/kotlin-js/src/commonMain/kotlin/js/typedarrays/Int8Array.kt#L11) what TypeScript decided to do, that is, add a generic type parameter:
```ts
interface Int8Array<TArrayBuffer extends ArrayBufferLike>
```

Because of issues such as [KT-61043](https://youtrack.jetbrains.com/issue/KT-61043) and [KT-60561](https://youtrack.jetbrains.com/issue/KT-60561), having type parameters makes actualization extremely difficult, if not impossible. This is why I had to replace kotlin-wrappers types with my own, or why I'm still using types from the stdlib.

PR #387, opened 2024-09-05
Read/write ByteString from/to ByteBuffer

### LINKED ISSUE #269: Add `ByteString` conversions for JVM `ByteBuffer`
In the same vibe as https://github.com/Kotlin/kotlinx-io/issues/266 and https://github.com/Kotlin/kotlinx-io/issues/268, it would be nice to have built-in conversions in `kotlinx-io-bytestring` for the JVM `ByteBuffer` type for a full multi-platform experience.

Examples:
```kotlin
/**
 * Exposes the contents of this [ByteString] as a read-only [ByteBuffer] without copying data.
 */
fun ByteString.asReadOnlyByteBuffer(): ByteBuffer = ByteBuffer.wrap(getBackingArrayReference()).asReadOnlyBuffer()

/**
 * Reads all remaining bytes in this [ByteBuffer] into a new [ByteString].
 */
fun ByteBuffer.readByteString(): ByteString = ByteString.wrap(readByteArray())

/**
 * Reads all remaining bytes in this [ByteBuffer] into a new [ByteArray].
 */
// private because that's not really the role of kotlinx-io-bytestring, more of the stdlib
private fun ByteBuffer.readByteArray(): ByteArray {
    val array = ByteArray(remaining())
    get(array)
    return array
}
```

Arguably the last one could be a stdlib function instead.

fzhinkin:
Closes #269

### DIFF
--- a/bytestring/jvm/src/ByteStringJvmExt.kt
+++ b/bytestring/jvm/src/ByteStringJvmExt.kt
@@ -5,6 +5,10 @@
 
 package kotlinx.io.bytestring
 
+import kotlinx.io.bytestring.unsafe.UnsafeByteStringApi
+import kotlinx.io.bytestring.unsafe.UnsafeByteStringOperations
+import java.nio.BufferOverflowException
+import java.nio.ByteBuffer
 import java.nio.charset.Charset
 
 /**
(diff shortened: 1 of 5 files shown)


bytestring/jvm/src/ByteStringJvmExt.kt
```diff
@@ -20,3 +24,103 @@ public fun ByteString.decodeToString(charset: Charset): String = getBackingArray
...
+ *
+ * This function does not update [ByteBuffer.position].
+ *
+ * @throws java.nio.ReadOnlyBufferException when [this] buffer is read-only
+ * @throws IndexOutOfBoundsException when [at] is negative, exceeds [ByteBuffer.limit], or
+ * [at] + [ByteString.size] exceeds [ByteBuffer.limit]
+ */
+public fun ByteBuffer.putByteString(at: Int, string: ByteString) {
+    checkIndexAndCapacity(at, string.size)
+    // Absolute get(byte[]) was added only in JDK 13
```
JakeWharton, line 109:
```suggestion
    // Absolute put(byte[]) was added only in JDK 16
```
https://docs.oracle.com/en%2Fjava%2Fjavase%2F21%2Fdocs%2Fapi%2F%2F/java.base/java/nio/ByteBuffer.html#put(int,java.nio.ByteBuffer,int,int)

shanshin (approved):

PR #390, opened 2024-09-11
Implement ByteString factory accepting UBytes

### LINKED ISSUE #389: ByteString: provide ByteString-factory overload accepting unsigned bytes
Currently, `ByteString`-factory accepts only signed bytes. Which is mildly annoying when it comes to writing code that initializes byte strings with some constants.

It would be nice to have an unsigned bytes overload.

```kotlin
// How it looks now
val magic = ByteString(0xcf.toByte(), 0xfa.toByte(), 0xed.toByte(), 0xfe.toByte())

// How should it look:
val magic = ByteString(0xcfu, 0xfau, 0xedu, 0xfeu)
```

fzhinkin:
Also, added a no-arg overload to resolve ambiguous `ByteString()` calls.

Fixes #389


bytestring/common/src/ByteString.kt
```diff
@@ -37,6 +38,23 @@ public fun ByteString(vararg bytes: Byte): ByteString = if (bytes.isEmpty()) {
     ByteString.wrap(bytes)
 }
 
+/**
+ * Wraps given [bytes] into a byte string.
+ *
```
shanshin, line 43: Is it worth clarifying (a seemingly obvious thing) how the written bytes will be read from ByteString (as `.toByte()`)?

fzhinkin: It definitely is

shanshin (approved):

Issue #394, opened 2024-09-23
Web crash `get_os().platform() == 'win32'`


raspberry-jenshen:
Hello!

Thank you for your cool package!

I faced issues when I converted my Kotlin code to the web platform (JS).

My code: 
```
internal class FileSystemImpl(
    private val fileSystem: KotlinFiles = SystemFileSystem,
) : FileSystem {
```

The crash:
`platform  undefined`

The crash happens when FileSystemImpl is created.

```
// JS code
var properties_initialized_FileSystemJs_kt_lg8f5i;
  function _init_properties_FileSystemJs_kt__4boaac() {
    if (!properties_initialized_FileSystemJs_kt_lg8f5i) {
      properties_initialized_FileSystemJs_kt_lg8f5i = true;
      SystemFileSystem = new SystemFileSystem$1();
      // crash is here
      isWindows = get_os().platform() == 'win32';
    }
  }
```

fzhinkin:
@raspberry-jenshen `SystemFileSystem` is not supported for JS/Browser. It's supported only for NodeJS environment.

raspberry-jenshen:
Oh, I see, please improve the documentation then

Couse, you have this target.

<img width="268" alt="Screenshot 2024-09-25 at 13 42 20" src="https://github.com/user-attachments/assets/4b175721-b3db-49cf-becb-20cf413f050c">

Issue #393, opened 2024-09-22
Add thread-safety comments to kdoc


vdshb:
I couldn't find any documentation about thread-safety of any class/interface in kotlinx-io. 

I am particularly interested in in thread safety of `Sink::write` from `SystemFileSystem.sink().buffered()`, but I believe it would be beneficial to add explicit remarks into kdoc about thread-safety of Sink, Source, Buffer, etc.

PR #395, opened 2024-09-23
Improve documentation

### LINKED ISSUE #392: Can you use peek to fill the buffer?
With a given source, if I run `source.peek().readInt()`, does this guarantee that 4 bytes will be read into the source's buffer?

If so, this should be documented in `Source.peek()`.

fzhinkin:
Explicitly documents thread-safety guarantees, PeekSource behavior and FileSystem support for JS-target.

Closes #392, #393, #394

### DIFF
--- a/core/common/src/RawSink.kt
+++ b/core/common/src/RawSink.kt
@@ -32,6 +32,11 @@ package kotlinx.io
  *
  * Implementors should abstain from throwing exceptions other than those that are documented for RawSink methods.
  *
+ * ### Thread-safety guarantees
+ *
+ * [RawSink] implementations are not required to be thread safe.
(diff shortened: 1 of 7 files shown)


core/common/src/RawSink.kt
```diff
@@ -32,6 +32,11 @@ package kotlinx.io
  *
  * Implementors should abstain from throwing exceptions other than those that are documented for RawSink methods.
  *
+ * ### Thread-safety guarantees
+ *
+ * [RawSink] implementations are not required to be thread safe.
+ * However, if an implementation provides some thread safety guarantees, it's recommended to explicitly document them.
```
qwwdfsad, line 38: We try to follow a more formal style -- `it's` -> `it is`

qwwdfsad (approved):

PR #399, opened 2024-10-11
Fix dokka templates

whyoleg:


build.gradle.kts
```diff
@@ -8,10 +8,10 @@ import kotlinx.validation.ExperimentalBCVApi
 
 plugins {
     id("kotlinx-io-publish") apply false
+    id("kotlinx-io-dokka")
```
whyoleg, line 11: Previously we haven't applied Dokka's convention plugin to root, and so it was not configured there

fzhinkin (approved):

PR #409, opened 2024-10-30
Publish Segment.size and Segment.remainingCapacity

### LINKED ISSUE #408: Unsafe API: make segment size and remaining capacity public
[Proposed Unsafe API design](https://github.com/Kotlin/kotlinx-io/issues/135#issuecomment-2125014081) explicitly mentioned availability `Segment.size` and `Segment.remainingCapacity` properties. 
Moreover, some of our samples use these properties. 
However, properties themselves remain internal.

Without having access to these properties it is quite hard to reason about number of bytes that could be read from or written to a segment using Unsafe API, thus we need to make them public.

fzhinkin:
Fixes #408

### DIFF
--- a/core/common/src/Segment.kt
+++ b/core/common/src/Segment.kt
@@ -345,14 +345,20 @@ public class Segment {
         limit += srcEndOffset - srcStartOffset
     }
 
-    @PublishedApi
-    @get:JvmSynthetic
-    internal val size: Int
+    /**
+     * The number of readable bytes contained in this segment.
+     *
+     * @sample kotlinx.io.samples.unsafe.UnsafeBufferOperationsSamples.readUleb128
+     */
+    public val size: Int
         get() = limit - pos
 
-    @PublishedApi
-    @get:JvmSynthetic
-    internal val remainingCapacity: Int
+    /**
+     * The number of bytes that could be written into this segment.
+     *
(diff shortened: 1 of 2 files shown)


core/common/src/Segment.kt
```diff
@@ -345,14 +345,20 @@ public class Segment {
...
 
-    @PublishedApi
-    @get:JvmSynthetic
-    internal val size: Int
+    /**
+     * The number of readable bytes contained in this segment.
+     *
+     * @sample kotlinx.io.samples.unsafe.UnsafeBufferOperationsSamples.readUleb128
+     */
+    public val size: Int
```
fzhinkin, line 353: Perhaps, it should be `Long`.

fzhinkin: But then both `pos` and `limit` have to be updated too.

fzhinkin: And the reason why we might need `Long` properties here is supporting MemorySegmented backed segments on JVM (and that's nowhere near).

JakeWharton: `MemorySegment`s can be sliced which should be a cheap operation. If one came in over 2^31 in length, it could be sliced into multiple segments.

qwwdfsad (approved):

shanshin (approved):

PR #410, opened 2024-10-30
Updated Dokka Gradle Plugin

fzhinkin:


build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-dokka.gradle.kts
```diff
@@ -21,13 +21,13 @@ tasks.withType<AbstractDokkaTask>().configureEach {
     )
 }
 
-tasks.withType<DokkaTaskPartial>().configureEach {
+dokka {
```
whyoleg, line 24: configuration from above should be also migrated to `dokka` extension.
e.g something like this:
```
dokka {
  pluginsConfiguration.html {
    templatesDir = rootDir.resolve("dokka-templates")
  }
}
```

fzhinkin: Thanks!

whyoleg (approved):
Nice!

Reminder: we will need to merge update to Dokka 2.0.0 synchronously with all other kotlinx.* because of HTML templates changes

whyoleg (commented):
JIC, the update to Dokka 2.0.0 was merged into master, so this PR should be rebased.

@fzhinkin, I also wanted to mention, that after this PR is merged, kotlinlang deploy configuration ([task](https://github.com/JetBrains/kotlin-web-site/blob/2ecbe9bc32d9c2e35237c1f9ac66266ad99018a0/.teamcity/builds/apiReferences/kotlinx/io/KotlinxIOBuildApiReference.kt#L28) and [pagesRoot](https://github.com/JetBrains/kotlin-web-site/blob/2ecbe9bc32d9c2e35237c1f9ac66266ad99018a0/.teamcity/builds/apiReferences/BuildApiPages.kt#L21)) should be updated to use new task name and output folder as it was changed in DGPv2.
An alternative will be to register a temporary task with the old name and change the `outputDirectory` for Dokka pulication output.

But I think that it might be better to just create a PR into `kotlin-web-site` repository with those changes (I could help with it if needed) and merge it just before the next release of `kotlinx-io`, or may be we can merge the PR right away, as for `kotlinx-io` it's configured to deploy only on tag anyway

Do you have any preferences here?

fzhinkin:
@whyoleg thanks! I think it's better to update the task on the web site side. I open the PR there: https://github.com/JetBrains/kotlin-web-site/pull/4839

PR #411, opened 2024-10-30
Move kotlinx.io.files package to a separate module

fzhinkin:


whyoleg:
JIC: there is [one usage](https://github.com/ktorio/ktor/blob/84daf56dcef0e42f768c389d20c58d1be4a94fa3/ktor-server/ktor-server-core/common/src/io/ktor/server/engine/ServerEngineUtils.kt#L7) of `kotlinx.io.files` in ktor `production` code (not in tests) blindly (not thinking about consequences) introduced by myself in https://github.com/ktorio/ktor/pull/4192
It was needed to not reimplement this from scratch (or copy from `kotlinx-io`) for js/wasm-js targets - previously it was just expect/actual
Probably it's better to drop this usage in `ktor` to avoid breaking changes? cc @e5l

fzhinkin:
In short term, the change should not break Ktor as long as a new artifact is added to dependencies.

fzhinkin:
One of the questions still opened for me is the name of a new module. It should include something like `-legacy` in its name. Otherwise, we'll have to put the future filesystem API into `kotlinx-io-filesystem-2`/`kotlinx-io-filesystem-new` module 😞 .

WDYT?

shanshin (approved):
LGTM

shanshin:
> One of the questions still opened for me is the name of a new module. It should include something like -legacy in its name. > Otherwise, we'll have to put the future filesystem API into kotlinx-io-filesystem-2/kotlinx-io-filesystem-new module

if we create a module with a different name, will we have two filesystem modules with a different name at the same time?
Why can't we use different package names?

fzhinkin:
@shanshin 

> if we create a module with a different name, will we have two filesystem modules with a different name at the same time?

Yes, we will.

> Why can't we use different package names?

It's always an option.

fzhinkin:
> One of the questions still opened for me is the name of a new module. It should include something like -legacy in its name. Otherwise, we'll have to put the future filesystem API into kotlinx-io-filesystem-2/kotlinx-io-filesystem-new module 😞 .

Ok, so we'll have to use different package names anyway and having both the legacy and a new API inside a single module will ease the migration as we'll be able to supply deprecation with a meaningful replacement expression.

fzhinkin:
Ok, the current approach with the module (and its name) looks fine: we'll add a new API into it, that'll unlock easier migration compared to two separate artifacts.

PR #414, opened 2024-11-15
Delegate Int|Short|Long.reverseBytes to Java Stdlib

fzhinkin:
Seems like JVM's JIT compiler can't pattern match bytes reversal, so it's worth explicitly delegating to Java Stdlib functions, that are intrinsics candidates.

For long values, the new implementation shows 20% perf improvement for reversal itself.

Functions depending on it, like readXXXLe or writeXXXLe also show some moderate performance improvement from the change:
- baseline (develop):
```
Benchmark                   (minGap)   Mode  Cnt          Score          Error  Units
IntLeBenchmark.benchmark         128  thrpt   15  387643797.206 ± 13843848.182  ops/s
LongLeBenchmark.benchmark        128  thrpt   15  314485977.158 ±  2249441.675  ops/s
ShortLeBenchmark.benchmark       128  thrpt   15  433907914.224 ±  6445437.883  ops/s
```

- optimized (this PR):
```
Benchmark                   (minGap)   Mode  Cnt          Score          Error  Units
IntLeBenchmark.benchmark         128  thrpt   15  432487978.191 ± 29393229.522  ops/s
LongLeBenchmark.benchmark        128  thrpt   15  336771673.575 ±  6093734.967  ops/s
ShortLeBenchmark.benchmark       128  thrpt   15  440459420.611 ± 11852873.667  ops/s
```

### DIFF
--- a/core/common/src/-Util.kt
+++ b/core/common/src/-Util.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 JetBrains s.r.o. and respective authors and developers.
+ * Copyright 2017-2024 JetBrains s.r.o. and respective authors and developers.
  * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENCE file.
  */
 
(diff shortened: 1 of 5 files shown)


shanshin (approved):
A good improvement for integers!

PR #417, opened 2024-11-18
Fix JVM toolchain version for smoke tests

fzhinkin:
By using lowest supported JVM toolchain (8), we should be able to catch an unintentional class file version change.

### DIFF
--- a/smoke-tests/src/test/resources/templates/jvm.build.gradle.kts
+++ b/smoke-tests/src/test/resources/templates/jvm.build.gradle.kts
@@ -24,3 +24,7 @@ dependencies {
     implementation("%DEPENDENCY%")
     testImplementation(kotlin("test"))
 }
+
+kotlin {
+    jvmToolchain(8)
+}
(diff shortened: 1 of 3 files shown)


smoke-tests/src/test/resources/templates/jvm.build.gradle.kts
```diff
@@ -24,3 +24,7 @@ dependencies {
     implementation("%DEPENDENCY%")
     testImplementation(kotlin("test"))
 }
+
+kotlin {
+    jvmToolchain(8)
```
shanshin, line 29: Shouldn't we add  `org.gradle.toolchains.foojay-resolver-convention` into settings file to resolve toolchains?

fzhinkin: Nice idea! I've been taken JDK-8 availability for granted.

shanshin (approved):

Issue #423, opened 2024-12-04
{Source,Buffer}.indexOf(ByteString, Int) should return start index for empty ByteString


fzhinkin:
For consistency with `CharSequence.indexOf` (which out functions are mimicking), `Source.indexOf(BS, Int)` and `Buffer.indexOf(BS, Int)` should return a start index (`0`, by default) when byte string to search is empty. 

Currently, both functions return `0` (disregard `startIndex` value).

JakeWharton:
I don't think you can get away with returning the start index without first requiring at least that much data to be loaded.

The `CharSequence` one doesn't _quite_ return the start index if the search string is empty and there's not enough data. Doing `"short".indexOf("", 10)` will return 5 and not 10.

I don't love the idea of returning whatever arbitrary maximum index is available from the `Source`/`Buffer` because it can be more non-determinstic as opposed to a fully-static `String`, so I think the only choice might be to require the data.

qwwdfsad:
On a side note -- could you please mention fuzzer in issues found by fuzzer? (so the final bounty list would be easier to compile)

fzhinkin:
After all, I convinced myself that `indexOf(empty byte string)` should behave the same way `CharSequence.indexOf("")`.
That means, `Buffer|Source.indexOf(ByteString, someIndex) == max(0, min(someIndex, buffer.size))`.

For `Source`, we'll try to fetch `max(0, someIndex)` bytes first.

PR #424, opened 2024-12-04
Refine Source|Buffer.indexOf(ByteBuffer) contracts

### LINKED ISSUE #422: Buffer.indexOf(ByteString) accepts negative indices
While `Source.indexOf(ByteString)` throws `IllegalArgumentException` on attempts to supply negative indices, `Buffer`'s overload accepts them and return some results.

Expected behavior: `Buffer.indexOf(ByteString(...), -1)` should throw IAE.

fzhinkin:
Closes #422

### DIFF
--- a/core/common/src/ByteStrings.kt
+++ b/core/common/src/ByteStrings.kt
@@ -10,6 +10,7 @@ import kotlinx.io.bytestring.isEmpty
 import kotlinx.io.bytestring.unsafe.UnsafeByteStringApi
 import kotlinx.io.bytestring.unsafe.UnsafeByteStringOperations
 import kotlinx.io.unsafe.UnsafeBufferOperations
+import kotlin.math.max
 import kotlin.math.min
 
 /**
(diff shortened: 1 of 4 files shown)


core/common/src/ByteStrings.kt
```diff
@@ -85,6 +85,8 @@ public fun Source.readByteString(byteCount: Int): ByteString {
  * expands the source's buffer as necessary until [byteString] is found. This reads an unbounded number of
  * bytes into the buffer. Returns `-1` if the stream is exhausted before the requested bytes are found.
  *
+ * For empty byte strings this function always return `0`.
```
fzhinkin, line 88: See #423. I'm not sure how to handle empty byte strings yet.

shanshin: It seems to me that this is really better than fetching the data, provided that we always ignore what has been fetched.

Maybe it should be explicitly written that in the case of an empty byte string, the data will never be read from the source?

fzhinkin: > Maybe it should be explicitly written that in the case of an empty byte string, the data will never be read from the source?

Then, it would be inconsistent with the `Buffer`'s "overload". These two functions already behave differently, but I'd like to keep them as close to each other as possible.

core/common/src/ByteStrings.kt
```diff
@@ -117,11 +117,24 @@ public fun Source.indexOf(byteString: ByteString, startIndex: Long = 0): Long {
...
+/**
+ * Returns the index of the first match for [byteString] in the buffer at or after [startIndex].
+ * Returns `-1` if the stream is exhausted before the requested bytes are found.
+ *
+ * For empty byte strings this function always return `0`.
+ *
+ * @param byteString the sequence of bytes to find within the buffer.
+ * @param startIndex the index into the buffer to start searching from.
+ *
+ * @throws IllegalArgumentException if [startIndex] is negative or if it is greater of equal to buffer's [Buffer.size].
```
fzhinkin, line 129: BTW, this behavior is also inconsistent with `CharSequence.indexOf`, which I'm considering maybe not as a paragon, but as something to be aligned with.

shanshin (approved):

PR #428, opened 2025-01-12
Introduce ByteString based on an existing ByteArray

Laxystem:
I implemented this using an existing class and an `extendable` boolean as I didn't want to break binary compatibility with existing usage of the `ByteStringBuilder` constructor, but it is possible to rewrite using an abstract class (should be ever-so-slightly-faster).

My use case: I have a game engine, and I'd like to avoid copying data around more than necessary. Usually, this would be a micro-optimization; But re-meshing everything every frame adds up quickly. This way, I can pass my existing vertex buffer, and an offset (that I can record using the buffer's own `size`), and simply override the changed vertex's data.

I did not add a `buildByteString` function as this is a lower-level use-case and should be treated as such; I expect most to not need the resulting `ByteString` at all, and simply use `apply` instead.

I would add a `DelicateIoApi` annotation, but I didn't seem to find one.

### DIFF
--- a/bytestring/common/src/ByteStringBuilder.kt
+++ b/bytestring/common/src/ByteStringBuilder.kt
@@ -17,44 +17,61 @@ import kotlin.math.max
  * The builder avoids additional copies and allocations when `size == capacity` when [toByteString] called,
  * thus it's recommended to specify expected [ByteString] size as `initialCapacity` when creating a builder.
  *
- * When a builder runs out of available capacity, a new byte sequence with extended capacity
- * will be allocated and previously written data will be copied into it.
- *
- * @param initialCapacity the initial size of an underlying byte sequence.
- *
  * @sample kotlinx.io.bytestring.samples.ByteStringSamples.builderSample
  * @sample kotlinx.io.bytestring.samples.ByteStringSamples.builderSampleWithoutAdditionalAllocs
  */
-public class ByteStringBuilder(initialCapacity: Int = 0) {
-    private var buffer = ByteArray(initialCapacity)
-    private var offset: Int = 0
+public class ByteStringBuilder private constructor(
+	private var buffer: ByteArray,
+	private var offset: Int,
+	private val extendable: Boolean
+) {
+	private val safeBuffer get() = if (extendable) buffer else buffer.copyOf()
 
     /**
      * The number of bytes being written to this builder.
      */
     public val size: Int
-        get() = offset
+        get() = if (extendable) offset else buffer.size
 
     /**
-     * The number of bytes this builder can store without an extension of an internal buffer.
+     * The number of bytes this builder can store without extending the internal buffer or throwing.
      */
     public val capacity: Int
         get() = buffer.size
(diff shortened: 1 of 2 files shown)


lppedd:
Maybe OT, but while we're talking about an easy way to initialize a `ByteString` with an existing array, I still can't find the same for a `Buffer`.

JakeWharton:
You can already wrap an existing `ByteArray` into a `ByteString` with [`UnsafeByteStringOperations.wrapUnsafe`](https://github.com/Kotlin/kotlinx-io/blob/ff2c5496ce98ec1623b2d0a07161f86b93906c1a/bytestring/common/src/unsafe/UnsafeByteStringOperations.kt#L24-L30).

> Maybe OT, but while we're talking about an easy way to initialize a `ByteString` with an existing array, I still can't find the same for a `Buffer`.

There is no way to give a `ByteArray` to a `Buffer` directly. A `Buffer` is comprised of a series of `Segment` instances which own their backing `ByteArray`, and those segments are pooled for efficiency. It's not impossible to design a system where you give a `ByteArray` which creates a `Segment` that is also never pooled, but I think it is unclear whether this is useful enough in contrast to some of the other APIs. We haven't needed such an API in Okio, for example.

What you can do, is ask a `Buffer` for the `ByteArray`s which back its `Segment`s to directly write to using [`UnsafeBufferOperations.writeToTail`](https://github.com/Kotlin/kotlinx-io/blob/ff2c5496ce98ec1623b2d0a07161f86b93906c1a/core/common/src/unsafe/UnsafeBufferOperations.kt#L154-L190). This is similar in capability to Okio's `UnsafeCursor` API.

fzhinkin:
@lppedd in addition to what Jake wrote, you can also use [UnsafeBufferOperations.moveToTail](https://github.com/Kotlin/kotlinx-io/blob/ff2c5496ce98ec1623b2d0a07161f86b93906c1a/core/common/src/unsafe/UnsafeBufferOperations.kt#L42) to gave ownership over an array to the buffer, which, to some extent, should be the same as initializing a buffer with an array.

fzhinkin:
@Laxystem one of the most crucial parts of the ByteString's contract is its immutability. The approach proposed in this PR let users to accidentaly break the contract much easily.

There is existing API that raises the same concern, however, it is ringing all the bells to ensure that nobody will break the contract unintentionally (and, hopefully, won't break it at all):
> You can already wrap an existing ByteArray into a ByteString with [UnsafeByteStringOperations.wrapUnsafe](https://github.com/Kotlin/kotlinx-io/blob/ff2c5496ce98ec1623b2d0a07161f86b93906c1a/bytestring/common/src/unsafe/UnsafeByteStringOperations.kt#L24-L30).

If some "unsafe" API is required, (most likely) the only place for it would be `UnsafeByteStringOperations`, not the regular public API.

I would love to read more details about a problem you're trying to solve. Perhaps, we'll figure out how to solve your problem without making the builder less "safe".

PR #432, opened 2025-02-12
Update to Dokka 2.0.0

whyoleg:


fzhinkin:
Should we merge https://github.com/Kotlin/kotlinx-io/pull/410 instead?

whyoleg:
> Should we merge https://github.com/Kotlin/kotlinx-io/pull/410 instead?

Not really, it's better to merge upgrade to DGPv2 separately, as it will require additional update of tasks which are executed during kotlinlang build (old `dokkaHtmlMultiModule` vs new `dokkaGenerate`).
It's better to make synchronous update with update of templates as simple as possible, and afterwards every project will be able to update DGPv2 + website configuration separately

fzhinkin (approved):

whyoleg:
(please don't delete `whyoleg/dokka2-sync` branch for some time until we finalize deployment changes)

PR #460, opened 2025-07-07
Support a new publication scheme

fzhinkin:


build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-publish.gradle.kts
```diff
@@ -68,22 +68,12 @@ fun MavenPublication.mavenCentralArtifacts(project: Project, sources: SourceDire
...
-    } else {
-        URI("https://oss.sonatype.org/service/local/staging/deployByRepositoryId/$repositoryId")
-    }
-}
-
-fun RepositoryHandler.configureMavenPublication( project: Project) {
+fun RepositoryHandler.configureMavenPublication(project: Project) {
     maven {
-        url = mavenRepositoryUri()
+        url = uri(project.getSensitiveProperty("libs.repo.url")!!)
```
shanshin, line 73:
```suggestion
        url = uri(project.getSensitiveProperty("libs.repo.url") ?: error("libs.repo.url is not set") )
```

fzhinkin: Updated it to simply ignore repository setup if the url is not provided (which is always the case, except running our CI tasks)

shanshin (approved):

PR #462, opened 2025-07-07
Update api reference url

### LINKED ISSUE #461: Published Dokka: Missing package list
When I configure my library's dokka with:

```kt
dokka {
  dokkaSourceSets {
    configureEach {
      externalDocumentationLinks {
        create("kotlinx-io") {
          url("https://kotlin.github.io/kotlinx-io/")
        }
      }
    }
  }
}
```

I see:

```
Failed to download package-list from https://kotlin.github.io/kotlinx-io/package-list, this might suggest that remote resource is not available, module is empty or dokka output got corrupted
```

And sure enough, looking at the gh-pages branch with dokka output, I don't see any package-list available: https://github.com/Kotlin/kotlinx-io/tree/gh-pages

sargunv:
resolves #461


fzhinkin (approved):
Thanks for fixing it!

PR #481, opened 2025-11-19
Fix loadModule for wasmJs

alex28sh:
…o be present there.

It worked previously, but recent changes in K/Wasm (that'll land in 2.3.20) triggered the following error:

> Cannot determine intended module format because both require() and top-level await are present.
> If the code is intended to be CommonJS, wrap await in an async function.
> If the code is intended to be an ES module, replace require() with import.

This change provides a workaround for the issue.

[1] https://kotlinlang.org/docs/wasm-js-interop.html#kotlin-wasm-and-kotlin-js-interoperability-differences

### DIFF
--- a/core/wasmJs/src/node/nodeModulesWasmJs.kt
+++ b/core/wasmJs/src/node/nodeModulesWasmJs.kt
@@ -5,24 +5,36 @@
 
 package kotlinx.io.node
 
+@JsFun("""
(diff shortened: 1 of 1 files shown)


alex28sh:
> @alex28sh please rebase the branch onto `develop`.

@fzhinkin done

Issue #463, opened 2025-07-24
FileMetadata with last modification information


claudiofelber:
Please add file timestamp information to the FileMetadata class. okio's FileMetadata provides the following three properties:

```Kotlin
val createdAtMillis: Long
val lastModifiedAtMillis: Long
val lastAccessedAtMillis: Long
```

This or something comparable would be really helpful. File timestamp information is required for a lot of use cases, such as determining whether a cached file needs to be updated or not.

ruben-caceiro-digiwest:
Any update on this?

Issue #297, opened 2024-04-22
Replace required reason APIs with equipment Foundations APIs on Apple targets


fzhinkin:
Starting from May 1st, 2024, developers have to explicitly describe why they are using some APIs for apps published in AppStore, apps won't be accepted (see [this](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api) for more info). 

One such API is `stat,` which we use to gather file metadata. It should be replaced with NSFileManager calls.

PR #488, opened 2025-12-27
FileMetadata with last modification information #463

rcaceiro:
Add support for file creation and modification timestamps in `FileMetadata` across platforms

This pull request closes issue #463

### DIFF
--- a/core/apple/src/files/FileSystemApple.kt
+++ b/core/apple/src/files/FileSystemApple.kt
@@ -10,6 +10,8 @@ import kotlinx.cinterop.*
 import kotlinx.io.IOException
 import platform.Foundation.*
 import platform.posix.*
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
 
 
 internal actual fun atomicMoveImpl(source: Path, destination: Path) {
@@ -53,14 +55,35 @@ internal actual fun realpathImpl(path: String): String {
     }
 }
 
+@OptIn(ExperimentalTime::class)
 internal actual fun metadataOrNullImpl(path: Path): FileMetadata? {
     val attributes = NSFileManager.defaultManager().fileAttributesAtPath(path.path, traverseLink = true) ?: return null
     val fileType = attributes[NSFileType] as String
     val isFile = fileType == NSFileTypeRegular
     val isDir = fileType == NSFileTypeDirectory
+    val createdAt = attributes[NSFileCreationDate]
+        .let {
+            it as? NSDate
+        }
+        ?.let(NSDate::toInstant)
+    val updatedAt = attributes[NSFileModificationDate]
+        .let {
+            it as? NSDate
+        }
+        ?.let(NSDate::toInstant)
     return FileMetadata(
         isRegularFile = isFile,
         isDirectory = isDir,
-        size = if (isFile) attributes[NSFileSize] as Long else -1
+        size = if (isFile) attributes[NSFileSize] as Long else -1,
+        createdAt = createdAt,
+        updatedAt = updatedAt,
     )
+}
+
+@OptIn(ExperimentalTime::class)
+private fun NSDate.toInstant(): Instant {
+    val epoch = this.timeIntervalSince1970
+    val seconds = epoch.toLong()
+    val nanos = (epoch - seconds).times(1e9).toLong()
+    return Instant.fromEpochSeconds(seconds, nanos)
 }
--- a/core/apple/src/files/FileSystemApple.kt
+++ b/core/apple/src/files/FileSystemApple.kt
@@ -74,14 +74,14 @@ internal actual fun metadataOrNullImpl(path: Path): FileMetadata? {
         isDirectory = isDir,
         size = if (isFile) attributes[NSFileSize] as Long else -1,
         createdAt = createdAt,
-        updatedAt = updatedAt,
+        lastModificationTime = updatedAt,
     )
 }
 
 private fun NSDate.toInstant(): Instant {
     val nanosInSecond = 1_000_000_000L
     val epochSeconds = this.timeIntervalSince1970
-    val epochSecondsTruncated = epoch.toLong()
+    val epochSecondsTruncated = epochSeconds.toLong()
     val nanos = (epochSeconds - epochSecondsTruncated).times(nanosInSecond).toLong()
     return Instant.fromEpochSeconds(epochSecondsTruncated, nanos)
 }
--- a/core/apple/src/files/FileSystemApple.kt
+++ b/core/apple/src/files/FileSystemApple.kt
@@ -79,8 +79,9 @@ internal actual fun metadataOrNullImpl(path: Path): FileMetadata? {
 }
 
 private fun NSDate.toInstant(): Instant {
-    val epoch = this.timeIntervalSince1970
-    val seconds = epoch.toLong()
-    val nanos = (epoch - seconds).times(1e9).toLong()
-    return Instant.fromEpochSeconds(seconds, nanos)
+    val nanosInSecond = 1_000_000_000L
+    val epochSeconds = this.timeIntervalSince1970
+    val epochSecondsTruncated = epoch.toLong()
+    val nanos = (epochSeconds - epochSecondsTruncated).times(nanosInSecond).toLong()
+    return Instant.fromEpochSeconds(epochSecondsTruncated, nanos)
 }
--- a/core/apple/src/files/FileSystemApple.kt
+++ b/core/apple/src/files/FileSystemApple.kt
@@ -10,7 +10,6 @@ import kotlinx.cinterop.*
 import kotlinx.io.IOException
 import platform.Foundation.*
 import platform.posix.*
-import kotlin.time.ExperimentalTime
 import kotlin.time.Instant
 
 
@@ -55,7 +54,6 @@ internal actual fun realpathImpl(path: String): String {
     }
 }
 
-@OptIn(ExperimentalTime::class)
 internal actual fun metadataOrNullImpl(path: Path): FileMetadata? {
     val attributes = NSFileManager.defaultManager().fileAttributesAtPath(path.path, traverseLink = true) ?: return null
     val fileType = attributes[NSFileType] as String
@@ -80,7 +78,6 @@ internal actual fun metadataOrNullImpl(path: Path): FileMetadata? {
     )
 }
 
-@OptIn(ExperimentalTime::class)
 private fun NSDate.toInstant(): Instant {
     val epoch = this.timeIntervalSince1970
     val seconds = epoch.toLong()
--- a/core/apple/src/files/FileSystemApple.kt
+++ b/core/apple/src/files/FileSystemApple.kt
@@ -74,7 +74,7 @@ internal actual fun metadataOrNullImpl(path: Path): FileMetadata? {
         isDirectory = isDir,
         size = if (isFile) attributes[NSFileSize] as Long else -1,
         createdAt = createdAt,
-        lastModificationTime = updatedAt,
+        lastModifiedAt = updatedAt,
     )
 }
 
--- a/core/common/src/files/FileSystem.kt
+++ b/core/common/src/files/FileSystem.kt
@@ -8,6 +8,8 @@ package kotlinx.io.files
 import kotlinx.io.IOException
 import kotlinx.io.RawSink
 import kotlinx.io.RawSource
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
 
 /**
  * An interface providing basic operations on a filesystem, such as reading and writing files,
@@ -192,6 +194,7 @@ public expect val SystemTemporaryDirectory: Path
  * Represents information about a file or directory obtainable from a filesystem like
  * a type of filesystem node (is it a directory or a file?), its size, and so on.
  */
+@OptIn(ExperimentalTime::class)
 public class FileMetadata(
     /**
      * Flag indicating that the metadata was retrieved for a regular file.
@@ -204,7 +207,15 @@ public class FileMetadata(
     /**
      * File size. Defined only for regular files, for other filesystem entities it will be `-1`.
      */
-    public val size: Long = 0L
+    public val size: Long = 0L,
+    /**
+     * Time when the file was created, or `null` if the filesystem does not support it.
+     */
+    public val createdAt: Instant? = null,
+    /**
+     * Time when the file was last modified, or `null` if the filesystem does not support it.
+     */
+    public val updatedAt: Instant? = null,
 )
 
 /**
--- a/core/common/src/files/FileSystem.kt
+++ b/core/common/src/files/FileSystem.kt
@@ -8,7 +8,6 @@ package kotlinx.io.files
 import kotlinx.io.IOException
 import kotlinx.io.RawSink
 import kotlinx.io.RawSource
-import kotlin.time.ExperimentalTime
 import kotlin.time.Instant
 
 /**
@@ -194,7 +193,6 @@ public expect val SystemTemporaryDirectory: Path
  * Represents information about a file or directory obtainable from a filesystem like
  * a type of filesystem node (is it a directory or a file?), its size, and so on.
  */
-@OptIn(ExperimentalTime::class)
 public class FileMetadata(
     /**
      * Flag indicating that the metadata was retrieved for a regular file.
--- a/core/common/src/files/FileSystem.kt
+++ b/core/common/src/files/FileSystem.kt
@@ -213,7 +213,7 @@ public class FileMetadata(
     /**
      * Time when the file was last modified, or `null` if the filesystem does not support it.
      */
-    public val updatedAt: Instant? = null,
+    public val lastModificationTime: Instant? = null,
 )
 
 /**
--- a/core/common/src/files/FileSystem.kt
+++ b/core/common/src/files/FileSystem.kt
@@ -213,7 +213,7 @@ public class FileMetadata(
     /**
      * Time when the file was last modified, or `null` if the filesystem does not support it.
      */
-    public val lastModificationTime: Instant? = null,
+    public val lastModifiedAt: Instant? = null,
 )
 
 /**
--- a/core/common/test/files/SmokeFileTest.kt
+++ b/core/common/test/files/SmokeFileTest.kt
@@ -7,7 +7,12 @@ package kotlinx.io.files
 
 import kotlinx.io.*
 import kotlin.test.*
+import kotlin.time.Clock
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
 
+@OptIn(ExperimentalTime::class)
 class SmokeFileTest {
     private val files: MutableList<Path> = arrayListOf()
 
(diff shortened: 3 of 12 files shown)


core/common/src/files/FileSystem.kt
```diff
@@ -192,6 +194,7 @@ public expect val SystemTemporaryDirectory: Path
  * Represents information about a file or directory obtainable from a filesystem like
  * a type of filesystem node (is it a directory or a file?), its size, and so on.
  */
+@OptIn(ExperimentalTime::class)
```
fzhinkin: Corresponding properties has to be annotated with `@ExperimentalTime` instead, but either way having it in the public API is undesirable. So it makes more sense to wait until the Kotlin version is updated to 2.3.

rcaceiro:
@fzhinkin 

Yes I can expose it no problem, I focused in add something that is needed.
Related to the experimental yes I used because the other one is deprecated ... I think as you know in 2.3 is stable and doing work for something that we have guarantee that will be removed is not a very good use of time... if as a maintainer you should update? 
yes of course, but is not a me decision ... I don't know your roadmap and if is a good time to do it

fzhinkin:
@rcaceiro, `develop` branch was updated to Kotlin 2.3, so now you rebase your branch and get rid of ExperimentalTime

rcaceiro:
@fzhinkin  Ok, its done

MohammedKHC:
Related: https://github.com/square/okio/pull/1772 was merged recently,  it adds creation time support for Linux and macOs.

core/jvm/src/files/FileSystemJvm.kt
```diff
@@ -84,8 +88,15 @@ public actual val SystemFileSystem: FileSystem = object : SystemFileSystemImpl()
...
-        return FileMetadata(path.file.isFile, path.file.isDirectory,
-            if (path.file.isFile) path.file.length() else -1L)
+        val attributes = Files.readAttributes(path.file.toPath(), BasicFileAttributes::class.java)
+
+        return FileMetadata(
+            isRegularFile = path.file.isFile,
+            isDirectory = path.file.isDirectory,
+            size = if (path.file.isFile) path.file.length() else -1L,
+            createdAt = attributes.creationTime().toInstant().toKotlinInstant(),
+            updatedAt = path.file.lastModified().let(Instant::fromEpochMilliseconds),
```
fzhinkin: Is there a reason to fetch information about the last modified time separately? Why can't it be taken from `attributes`?

core/nativeNonApple/src/files/FileSystemNativeNonApple.kt
```diff
@@ -27,7 +27,9 @@ internal actual fun metadataOrNullImpl(path: Path): FileMetadata? {
         return FileMetadata(
             isRegularFile = isFile,
             isDirectory = (mode and S_IFMT) == S_IFDIR,
-            if (isFile) struct_stat.st_size.toLong() else -1L
+            size = if (isFile) struct_stat.st_size.toLong() else -1L,
+            createdAt = null,
```
fzhinkin: All three fields corresponding to b/c/m/a-times are accessible on individual platforms, but they have different names.

One of the possible solutions is to define extension expect properties returning `Instant` corresponding to one of those times and actualize them for different platforms.

Like
```
internal expect val stat.mtime: Instant

// actualized by

internal actual val stat.mtime: Instant
    get() = this.st_mtim.let {
        Instant.fromEpochSeconds(it.tv_sec.convert(), it.tv_nsec)
    }
```

core/nodeFilesystemShared/src/node/fs.kt
```diff
@@ -70,7 +70,10 @@ internal external interface Fs {
  * Partial declaration of a class mirroring [node:fs.Stats](https://nodejs.org/api/fs.html#class-fsstats)
  */
 internal external interface Stats {
+    val atimeMs: Double
+    val ctimeMs: Double
     val mode: Int
+    val mtimeMs: Double
```
fzhinkin, line 75:
```suggestion
    val atimeMs: Double
    val ctimeMs: Double
    val mtimeMs: Double
    val mode: Int
```

core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -572,7 +575,12 @@ private class FileSource(private val fd: Fd) : RawSource {
     }
 }
 
-private data class InternalMetadata(val filetype: FileType, val filesize: Long)
+private data class InternalMetadata(
+    val createdAt: Instant?,
+    val filesize: Long,
+    val filetype: FileType,
+    val modifiedAt: Instant,
```
fzhinkin:
```suggestion
    val filesize: Long,
    val filetype: FileType,
    val createdAt: Instant?,
    val modifiedAt: Instant,
```

core/wasmWasi/src/files/FileSystemWasm.kt
```diff
@@ -594,7 +601,18 @@ private fun metadataOrNullInternal(rootFd: Fd, path: Path, followSymlinks: Boole
         if (res == Errno.noent || res == Errno.notcapable) return null
         if (res != Errno.success) throw IOException(res.description)
 
-        return InternalMetadata(filestat.filetype, filestat.filesize)
+        val modifiedAt = run {
+            val epoch = filestat.modificationTime * 1e-9
+            val seconds = epoch.toLong()
+            val nanos = (epoch - seconds) * 1e9
+            Instant.fromEpochSeconds(seconds, nanos.toLong())
```
fzhinkin: Why do you use floating-point arithmetics for the conversion?

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
```diff
@@ -80,7 +81,9 @@ public actual val SystemFileSystem: FileSystem = object : SystemFileSystemImpl()
             metadata = FileMetadata(
                 isRegularFile = isFile,
                 isDirectory = (mode and fs.constants.S_IFMT) == fs.constants.S_IFDIR,
-                if (isFile) stat.size.toLong() else -1L
+                size = if (isFile) stat.size.toLong() else -1L,
+                createdAt = stat.ctimeMs.toInstant(),
```
fzhinkin: Note that `ctimeMs` is not a creation time: https://nodejs.org/api/fs.html#statsctimems

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
```diff
@@ -128,3 +131,10 @@ public actual open class FileNotFoundException actual constructor(
     message: String?,
 ) : IOException(message)
 
+
+private fun Double.toInstant(): Instant {
+    val epoch = this * 0.001
+    val seconds = epoch.toLong()
+    val nanos = (epoch - seconds).times(1e9).toLong()
+    return Instant.fromEpochSeconds(seconds, nanos)
```
fzhinkin, line 140:
```suggestion
    return Instant.fromEpochMilliseconds(0L) + this.milliseconds
```

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
```diff
@@ -128,3 +131,10 @@ public actual open class FileNotFoundException actual constructor(
     message: String?,
 ) : IOException(message)
 
+
+private fun Double.toInstant(): Instant {
```
fzhinkin: Please use a more descriptive name. Like `toInstantFromEpochMillis`.

core/common/test/files/SmokeFileTest.kt
```diff
@@ -259,14 +262,22 @@ class SmokeFileTest {
...
     fun fileMetadata() {
+        val validateAt: (Instant?) -> Boolean = {
+            val now = Clock.System.now()
+            val margin = 5.seconds
+            it == null || (now.minus(margin) < it && now.plus(margin) > it)
+        }
         val path = createTempPath()
-        assertNull(SystemFileSystem.metadataOrNull(path))
+        val metadata = SystemFileSystem.metadataOrNull(path)
+        assertNull(metadata)
```
fzhinkin: Why?

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
fzhinkin: Or the current implementation gives better precision?

core/common/test/files/SmokeFileTest.kt
```diff
@@ -259,14 +262,22 @@ class SmokeFileTest {
 
     @Test
     fun fileMetadata() {
+        val validateAt: (Instant?) -> Boolean = {
+            val now = Clock.System.now()
+            val margin = 5.seconds
+            it == null || (now.minus(margin) < it && now.plus(margin) > it)
```
fzhinkin: There's no guarantee the test won't hang for more than 5 seconds in between creating a file and reading its attributes.

It makes sense to create a file, read its metadata, sleep for some period of time exceeding the minimal precision of file timestamps, update the file, read metadata and check that:
- modification time increased (if it was present)
- creation time remained the same (if it was present)

core/common/src/files/FileSystem.kt
```diff
@@ -204,7 +207,15 @@ public class FileMetadata(
...
-    public val size: Long = 0L
+    public val size: Long = 0L,
+    /**
+     * Time when the file was created, or `null` if the filesystem does not support it.
+     */
+    public val createdAt: Instant? = null,
+    /**
+     * Time when the file was last modified, or `null` if the filesystem does not support it.
+     */
+    public val updatedAt: Instant? = null,
```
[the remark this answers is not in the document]
fzhinkin: @rcaceiro, The attribute is named either `modification time` or `last modification time` elsewhere. It make sense to name it `lastModifiedAt` to be consistent with both docs ("Time when the file was **last modified**") and conventional names in other APIs.

core/common/src/files/FileSystem.kt
```diff
@@ -204,7 +207,15 @@ public class FileMetadata(
...
 )
```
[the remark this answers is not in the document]
fzhinkin: @rcaceiro, It seems like the attribute is available in every other platform. So please add it to the metadata too.

core/apple/src/files/FileSystemApple.kt
```diff
@@ -58,9 +59,28 @@ internal actual fun metadataOrNullImpl(path: Path): FileMetadata? {
...
+        createdAt = createdAt,
+        updatedAt = updatedAt,
     )
 }
+
+private fun NSDate.toInstant(): Instant {
+    val epoch = this.timeIntervalSince1970
+    val seconds = epoch.toLong()
+    val nanos = (epoch - seconds).times(1e9).toLong()
+    return Instant.fromEpochSeconds(seconds, nanos)
```
fzhinkin:
```suggestion
    val nanosInSecond = 1_000_000_000L
    val epochSeconds = this.timeIntervalSince1970
    val epochSecondsTruncated = epoch.toLong()
    val nanos = (epochSeconds - epochSecondsTruncated).times(nanosInSecond).toLong()
    return Instant.fromEpochSeconds(epochSecondsTruncated, nanos)
```

Or you can use the trick with `Duration`, for simplicity.

Either way, please declare a single function used by both node-fs-shared and Apple source sets.

fzhinkin (changes requested):
@rcaceiro please make sure that all checks pass locally. You can refer https://github.com/Kotlin/kotlinx-io/blob/master/CONTRIBUTING.md#building for details.

core/nativeNonApple/src/files/FileSystemNativeNonApple.kt
rcaceiro: Hi this ``` st_mtim ```  don't exist on stat

```

public expect final class stat public constructor(rawPtr: kotlin.native.internal.NativePtr /* from: kotlinx.cinterop.NativePtr */) : kotlinx.cinterop.CStructVar {
    @kotlin.Deprecated public expect companion object : kotlinx.cinterop.CStructVar.Type {
    }

    @kotlinx.cinterop.UnsafeNumber public expect final var st_mode: kotlin.UShort /* from: platform.posix.uint_fast16_t */ /* compiled code */

    @kotlinx.cinterop.UnsafeNumber public expect final var st_dev: kotlin.UInt /* compiled code */

    @kotlinx.cinterop.UnsafeNumber public expect final var st_ino: kotlin.UShort /* from: platform.posix.uint_fast16_t */ /* compiled code */

    @kotlinx.cinterop.UnsafeNumber public expect final var st_size: kotlin.Int /* compiled code */

    @kotlinx.cinterop.UnsafeNumber public expect final var st_rdev: kotlin.UInt /* compiled code */
}
```

core/jvm/src/files/FileSystemJvm.kt
rcaceiro: There was, the code is made by me, and I know that updated time was possible to get from file... then I had to find out were I are the created, and I don't changed

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
rcaceiro: Done

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
rcaceiro: Yes the current one has better precision

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
rcaceiro: Yes my bad, but I dont have the birthtime so has to ne null
```
internal external interface Stats {
    val atimeMs: Double
    val ctimeMs: Double
    val mtimeMs: Double
    val mode: Int
    val size: Int
    fun isDirectory(): Boolean
}
```

core/wasmWasi/src/files/FileSystemWasm.kt
rcaceiro: So, the modification time is nanos, so I have two options to get the seconds and the nanos seperatly:

1.  using floating multiplication, that is faster, and the one that I used for that
2. using division with integers, that are slower, but what i see im my research mutch more accureate

So I change to division for accuracy

core/common/test/files/SmokeFileTest.kt
rcaceiro: I add the time that took from the filsystem to do its job this way if the filesystem hangs it not fail

core/nativeNonApple/src/files/FileSystemNativeNonApple.kt
```diff
@@ -11,7 +11,7 @@ import platform.posix.*
 
 @OptIn(ExperimentalForeignApi::class)
 public actual val SystemTemporaryDirectory: Path
-    get() = Path(getenv("TMPDIR")?.toKString() ?: getenv("TMP")?.toKString() ?: "")
+    get() = Path(getenv("TMPDIR")?.toKString() ?: getenv("TMP")?.toKString() ?: "/tmp")
```
fzhinkin: It seems unrelated to extending file metadata.

core/nativeNonApple/src/files/FileSystemNativeNonApple.kt
fzhinkin: These are null by default, aren't they?

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
fzhinkin: Thanks!

core/nodeFilesystemShared/src/files/FileSystemNodeJs.kt
fzhinkin: Got you, thanks!

core/common/src/files/FileSystem.kt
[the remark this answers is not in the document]
fzhinkin: @rcaceiro now two new properties have inconsistent names: `createdAt` and `lastModificationTime`. It is always nice to use consistent naming, so either they are `createdAt` and `lastModifiedAt`, or `creationTime` and `lastModificationTime`. Personally, I'd vote for the former, `createdAt` and `lastModifiedAt`.

core/nativeNonApple/src/files/FileSystemNativeNonApple.kt
fzhinkin: In the mingw sourceset there is `st_mtime`, in the unix sourceset there's `st_mtim`.
You can declare an expect function extracting the modification time from stat struct in the nativeNonAppleSource set and actualize it in unix and mingw source set.

With creation time things are much worse, and it could be considered unavailable (although, on linux it could be obtained using `statx`, but there's no reasonable way to access `statx` from Kotlin).

core/common/test/files/SmokeFileTest.kt
```diff
@@ -259,26 +264,46 @@ class SmokeFileTest {
...
+                }
         }
 
         try {
             val fileMetadata = SystemFileSystem.metadataOrNull(filePath)
             assertNotNull(fileMetadata)
             assertFalse(fileMetadata.isDirectory)
             assertTrue(fileMetadata.isRegularFile)
+            assertTrue(validateAt(fileMetadata.createdAt, time))
+            assertTrue(validateAt(fileMetadata.lastModificationTime, time))
```
fzhinkin: Please also check that if something will be written to the filepath after this check, then the modification time will be increased, but the creation time will remain the same (assuming both times are available).

And it's worth testing that at this point they are equal to each other (they should be, right?).

fzhinkin (changes requested):

core/common/src/files/FileSystem.kt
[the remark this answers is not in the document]
rcaceiro: Yes I agree how ever in kotlin native the struct stat don't expose both create nor modified time

core/common/test/files/SmokeFileTest.kt
rcaceiro: I’ve completed the tests, but I’ve noted a discrepancy on JVM Linux: createdAt matches lastModifiedAt.

According to the documentation, if the file system doesn't support a specific creation timestamp, the JVM returns a default value—typically the last-modified time. Since this is expected behavior for certain Linux environments, I propose we skip this specific check on the JVM to allow the rest of the suite to pass.

How would you like me to proceed?

MohammedKHC:
@fzhinkin @rcaceiro
I was looking into why kotlinx-io uses `NSFileManager` instead of `stat` on Apple platforms 
and found #297, which explains that `stat` is restricted by the App Store.

However, now that timestamps have been added to the kotlinx-io API, I think this is worth 
revisiting — `FileMetadata` exposes timestamps which are also subject to App Store restrictions, 
so the switch from `stat` to `NSFileManager` doesn't actually help if timestamps are being read.

What's the thinking here? Is there maybe a way to only retrieve timestamps on demand on iOS?

Related:
- https://developer.apple.com/documentation/bundleresources/app-privacy-configuration/nsprivacyaccessedapitypes/nsprivacyaccessedapitype#possibleValues
- https://developer.apple.com/documentation/Foundation/FileAttributeKey/creationDate#Discussion

Edit
https://kotlinlang.org/docs/multiplatform/multiplatform-privacy-manifest.html#compose-multiplatform seems  also related.

core/common/test/files/SmokeFileTest.kt
fzhinkin: > I propose we skip this specific check on the JVM to allow the rest of the suite to pass.

Let's proceed with that. I don't really like the idea of returning some creation time on JVM when it is actually unknown, but given the NIO API I don't see a reasonable workaround for that.

fzhinkin:
@MohammedKHC, thanks for pointing to it, I totally forgot that timestamp retrieval is a restricted operation. :/

I don't have a quick answer what to do, and need to think about it.

core/common/test/files/SmokeFileTest.kt
rcaceiro: Done

core/nativeNonApple/src/files/FileSystemNativeNonApple.kt
rcaceiro: Reverted

rcaceiro:
I think that it all

mipastgt:
Any news on this? How can you actually use this library in a serious project without proper file metadata?

PR #489, opened 2026-01-05
Add Implementation-* attributes to the Jar file manifest

### LINKED ISSUE #492: Fill Implementation-Title/Version/Vendor attributes in JAR's Manifest
Having this information will help with improving DX in IDE.

See https://youtrack.jetbrains.com/issue/KT-83393

fzhinkin:
Fixes #492

### DIFF
--- a/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
+++ b/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
@@ -119,7 +119,12 @@ kotlin {
     tasks {
         val jvmJar by existing(Jar::class) {
             manifest {
-                attributes("Multi-Release" to true)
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
```diff
@@ -119,7 +119,12 @@ kotlin {
     tasks {
         val jvmJar by existing(Jar::class) {
             manifest {
-                attributes("Multi-Release" to true)
+                attributes(
+                    "Multi-Release" to true,
+                    "Implementation-Vendor" to "JetBrains",
+                    "Implementation-Title" to project.name,
```
shanshin: I'm not really aware of how the IDE uses this field. Wouldn't the short name "core" cause confusion?

PR #498, opened 2026-02-08
Add Source/Sink wrappers for kotlinx-browser's ArrayBuffer

cedrickcooke:
Placed this in an `integration` folder like Okio has, to avoid adding a dependency to `core`.


JakeWharton:
If these are going to be supported as sink and source targets, I think they should be in the core. The kotlinx-browser dependency isn't required, we can define our own externals to avoid having a dependency.

integration/kotlinx-io-browser/web/src/ArrayBufferSink.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlinx.io.RawSink
+import org.khronos.webgl.ArrayBuffer
+import org.khronos.webgl.Int8Array
+
+/**
+ * A [RawSink] that writes data to an [ArrayBuffer].
+ *
+ * @param arrayBuffer the buffer to write to.
+ */
+public class ArrayBufferSink(
```
JakeWharton: I don't think either of these types should be public. If we're following the library and Kotlin conventions they should be in this shape:
```kotlin
fun ArrayBuffer.asSink(): RawSink
fun ArrayBuffer.asSource(): RawSource
```
It's probably also worth defining this for `UInt8Array` directly. Or possibly even _only_ for `UInt8Array`...

The corresponding Okio issue for this feature is https://github.com/square/okio/issues/689

cedrickcooke:
Done. PR/branch title are a bit out of date now, though 😆

core/web/src/external/ArrayBuffer.kt
```diff
@@ -0,0 +1,17 @@
+@file:OptIn(kotlin.js.ExperimentalWasmJsInterop::class)
+
+package kotlinx.io.external
+
+import kotlin.js.JsAny
+
+/** Interop type for JavaScript `ArrayBuffer`. */
+public external class ArrayBuffer : JsAny {
```
JakeWharton: Hmm I may have misled you by saying that this library could just copy the externals. Because normally that's what I do in libraries to avoid kotlinx-browser, but here these need to be public because the extensions hang off of them. That's... not great. I really don't like the whole kotlinx-browser approach. For `Int8Array` you can use Kotlin's normal `ByteArray`, but I'm not sure if `UInt8Array` is mapped to `UByteArray` or not. And there's no solution in the stdlib for `ArrayBuffer`. I will defer to the JetBrains people for what they think is the best approach.

PR #507, opened 2026-05-10
Improve native `FileSource` and `FileSink`

### LINKED ISSUE #438: Reading from tty file doesn't behave the same on JVM vs Native
My app is trying to read from a tty file like this:
```kotlin
  val source = SystemFileSystem.source(Path("/dev/ttyUSB0")).buffered()
  var count = 0
  while (true) {
    val b = source.readByte()
    count++
    println("Received '${Char(b.toInt())}' - count=$count")
  }
```

- On JVM, this will print single bytes as soon as they are available
- On Native, this will wait for 8k bytes before printing all of them together, then wait for the next 8k

### Reproducer
I've setup a project [here](https://github.com/BoD/repro-kotlinxio-tty).

To setup you can use `socat` to create 2 linked pseudo-tty, like this:
```
socat PTY,link=/tmp/source,raw,echo=0 PTY,link=/tmp/sink,raw,echo=0 &
```

### More info
- For regular files, no difference between Java/Native 
- I guess `fread` [here](https://github.com/Kotlin/kotlinx-io/blob/master/core/native/src/files/PathsNative.kt#L86) won't return partial results. Not sure if that's expected and/or intended.
- But Java's `InputStream.read` [here](https://github.com/Kotlin/kotlinx-io/blob/master/core/jvm/src/JvmCore.kt#L85) does.
- Some discussion about buffering in #375 may be related (I think stdin/stdout would probably behave the same as tty files?)

### LINKED ISSUE #493: Open files with O_CLOEXEC flag
`O_CLOEXEC` forces file descriptors closing on invocation of functions from the `exec`-family. There are a few reasons why it's a good idea ranging from resources leakage to improved security.

It's a good idea to use this flag when opening file sinks and sources.

Inspired by https://github.com/whyoleg/cryptography-kotlin/issues/138

MohammedKHC:
- Switch to unbuffered I/O (open/read/write)
- Use UnsafeBufferOperations for zero-copy segment access
- Enable `O_CLOEXEC` on Unix platforms

Closes #438, Closes #493

### DIFF
--- a/core/mingw/src/files/FileSystemMingw.kt
+++ b/core/mingw/src/files/FileSystemMingw.kt
@@ -62,4 +62,4 @@ internal actual fun realpathImpl(path: String): String {
 }
 
 internal actual val DefaultOpenFlags: Int
-    get() = _O_BINARY
+    get() = _O_BINARY or _O_NOINHERIT
--- a/core/mingw/src/files/FileSystemMingw.kt
+++ b/core/mingw/src/files/FileSystemMingw.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
  * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
  */
 
@@ -60,3 +60,6 @@ internal actual fun realpathImpl(path: String): String {
         return buffer.toKString()
     }
 }
+
+internal actual val DefaultOpenFlags: Int
+    get() = _O_BINARY
--- a/core/native/src/files/FileSystemNative.kt
+++ b/core/native/src/files/FileSystemNative.kt
@@ -1,18 +1,16 @@
 /*
- * Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
  * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
  */
 
 package kotlinx.io.files
 
-import kotlinx.cinterop.CPointer
 import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.cinterop.toKString
 import kotlinx.io.IOException
 import kotlinx.io.RawSink
 import kotlinx.io.RawSource
 import platform.posix.*
-import kotlin.experimental.ExperimentalNativeApi
 
 @OptIn(ExperimentalForeignApi::class)
 public actual val SystemFileSystem: FileSystem = object : SystemFileSystemImpl() {
@@ -73,20 +71,27 @@ public actual val SystemFileSystem: FileSystem = object : SystemFileSystemImpl()
     }
 
     override fun source(path: Path): RawSource {
-        val openFile: CPointer<FILE>? = fopen(path.path, "rb")
-        if (openFile == null) {
+        val openFd = open(path.path, O_RDONLY or DefaultOpenFlags)
+        if (openFd < 0) {
             if (errno == ENOENT) {
                 throw FileNotFoundException("File does not exist: $path")
             }
             throw IOException("Failed to open $path with ${strerror(errno)?.toKString()}")
         }
-        return FileSource(openFile)
+        return FileSource(openFd)
     }
 
     override fun sink(path: Path, append: Boolean): RawSink {
-        val openFile: CPointer<FILE> = fopen(path.path, if (append) "ab" else "wb")
-            ?: throw IOException("Failed to open $path with ${strerror(errno)?.toKString()}")
-        return FileSink(openFile)
+        val openFd = open(
+            path.path,
+            O_WRONLY or O_CREAT or DefaultOpenFlags
+                    or (if (append) O_APPEND else O_TRUNC),
+            PermissionDefault
+        )
+        if (openFd < 0) {
+            throw IOException("Failed to open $path with ${strerror(errno)?.toKString()}")
+        }
+        return FileSink(openFd)
     }
 
     override fun list(directory: Path): Collection<Path> {
@@ -118,8 +123,13 @@ public actual open class FileNotFoundException actual constructor(
     message: String?
 ) : IOException(message)
 
+// 666 in octal, rw for all (owner, group and others).
+internal const val PermissionDefault: UShort = 0b110110110u
+
 // 777 in octal, rwx for all (owner, group and others).
-internal const val PermissionAllowAll: UShort = 511u
+internal const val PermissionAllowAll: UShort = 0b111111111u
+
+internal expect val DefaultOpenFlags: Int
 
 internal expect class OpaqueDirEntry : AutoCloseable {
     fun readdir(): String?
--- a/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
+++ b/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts
@@ -105,6 +105,10 @@ kotlin {
                     group("mingw")
                     group("linux")
                 }
+                group("nativeNonMingw") {
+                    group("apple")
+                    group("unix")
+                }
             }
             group("nodeFilesystemShared") {
                 withJs()
--- a/core/native/src/files/PathsNative.kt
+++ b/core/native/src/files/PathsNative.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 JetBrains s.r.o. and respective authors and developers.
+ * Copyright 2017-2026 JetBrains s.r.o. and respective authors and developers.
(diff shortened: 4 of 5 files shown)


MohammedKHC:
For context on why we shouldn't use buffered IO
https://github.com/square/okio/issues/1807

core/mingw/src/files/FileSystemMingw.kt
```diff
@@ -60,3 +60,6 @@ internal actual fun realpathImpl(path: String): String {
         return buffer.toKString()
     }
 }
+
+internal actual val DefaultOpenFlags: Int
+    get() = _O_BINARY
```
fzhinkin: Does it make sense to use `O_BINARY` on other platform too?

Also, it seems like Window's equivalent of `O_CLOEXEC` is `O_NOINHERIT`.

core/native/src/files/FileSystemNative.kt
```diff
@@ -118,8 +123,13 @@ public actual open class FileNotFoundException actual constructor(
     message: String?
 ) : IOException(message)
 
+// 666 in octal, rw for all (owner, group and others).
+internal const val PermissionDefault: UShort = 0b110110110u
```
fzhinkin, line 127: `fopen` takes umask into account. We should probably do the same.

fzhinkin:
@MohammedKHC do you see any actual performance / footprint improvement caused by bypassing the additional bufferization layer?

core/mingw/src/files/FileSystemMingw.kt
MohammedKHC: `O_BINARY` is Windows-only — Windows text mode (the default mode) translates `\n` to `\r\n`, `O_BINARY` disables that. Linux/macOs has no such translation so the flag doesn't exist. https://stackoverflow.com/questions/2266992/no-o-binary-and-o-text-flags-in-linux

> Also, it seems like Window's equivalent of O_CLOEXEC is O_NOINHERIT.

Thanks for noticing this, I'll add `O_NOINHERIT` as well.

MohammedKHC:
Haven't benchmarked it, but bypassing stdio's buffering layer is consistent with the JVM implementation. It also works better with ttys and opens the door for reading/writing to stdout/stderr in the future.

core/native/src/files/FileSystemNative.kt
MohammedKHC: I don't think we need to do anything extra

From  [open(2)](https://man7.org/linux/man-pages/man2/open.2.html) man page

> The effective permissions are modified by the process's umask in the usual way: in the absence of a default ACL, the permissions of the created file are `(mode & ~umask)`.

MohammedKHC: And also that's how the jvm opens files (0666)
https://github.com/openjdk/jdk/blob/master/src/java.base/unix/native/libjava/io_util_md.c#L123

fzhinkin:
> Haven't benchmarked it, but bypassing stdio's buffering layer is consistent with the JVM implementation. It also works better with ttys and opens the door for reading/writing to stdout/stderr in the future.

@MohammedKHC, it makes perfect sense, but nonetheless it's worth double checking that performance actually improves.

MohammedKHC:
@fzhinkin 
I wrote a simple benchmark comparing the old `fopen` implementation (with zero segment copies) against the new implementation.

I’m not even sure the benchmark is entirely correct — this is actually the first benchmark I’ve written 😅

I tested this on Linux.

Here are the results with a 10 MB file:

```
native summary:
Benchmark                 Mode  Cnt  Score   Error    Units
NewFilesBenchmark.read   thrpt    5  66.433 ± 0.664  ops/sec
NewFilesBenchmark.write  thrpt    5  23.696 ± 1.178  ops/sec
OldFilesBenchmark.read   thrpt    5  76.506 ± 7.644  ops/sec
OldFilesBenchmark.write  thrpt    5  20.305 ± 0.594  ops/sec
````

And with a 512 MB file:

```
native summary:
Benchmark                 Mode  Cnt  Score   Error    Units
NewFilesBenchmark.read   thrpt    5  1.683 ± 0.052  ops/sec
NewFilesBenchmark.write  thrpt    5  0.490 ± 0.043  ops/sec
OldFilesBenchmark.read   thrpt    5  1.713 ± 0.016  ops/sec
OldFilesBenchmark.write  thrpt    5  0.481 ± 0.006  ops/sec
```

At least from these results, the larger-file benchmark makes the two implementations look effectively equivalent. The old implementation seems to slightly favor reads, while the new implementation consistently performs better for writes in both benchmarks.

<details>
<summary>Benchmark code</summary>

~~~kotlin
@State(Scope.Benchmark)
abstract class FilesBenchmark {
    private lateinit var tempFile: Path
    private lateinit var data: ByteArray

    @Param("10", "512")
    val fileSizeMb: Int = 0

    @Setup
    fun setup() {
        tempFile = Path(SystemTemporaryDirectory, "file.txt")
        data = Random.nextBytes(fileSizeMb * 1024 * 1024)
        SystemFileSystem.sink(tempFile).buffered().use {
            it.write(data)
        }
    }

    @TearDown
    fun teardown() {
        SystemFileSystem.delete(tempFile)
    }

    abstract fun source(path: Path): RawSource
    abstract fun sink(path: Path): RawSink

    @Benchmark
    fun read() {
        source(tempFile).buffered().use {
            it.readByteArray()
        }
    }

    @Benchmark
    fun write() {
        sink(tempFile).buffered().use {
            it.write(data)
        }
    }
}

@State(Scope.Benchmark)
open class NewFilesBenchmark : FilesBenchmark() {
    override fun source(path: Path) = SystemFileSystem.source(path)
    override fun sink(path: Path) = SystemFileSystem.sink(path)
}

@State(Scope.Benchmark)
open class OldFilesBenchmark : FilesBenchmark() {
    override fun source(path: Path): RawSource = FopenZeroCopyFileSource(path)
    override fun sink(path: Path): RawSink = FopenZeroCopyFileSink(path)
}

@OptIn(ExperimentalForeignApi::class, UnsafeIoApi::class)
internal class FopenZeroCopyFileSource(path: Path) : RawSource {
    private val file = fopen(path.toString(), "rb")!!
    private var closed = false

        val bytesRead = UnsafeBufferOperations.writeToTail(sink, 1) { bytes, start, end ->
            val toRead = minOf((end - start).toLong(), byteCount)
            bytes.usePinned {
                fread(it.addressOf(start), 1u, toRead.toULong(), file).toInt()
            }
        }.toLong()

        return when {
            bytesRead == byteCount -> bytesRead
            feof(file) != 0 -> if (bytesRead == 0L) -1L else bytesRead
            ferror(file) != 0 -> throw IOException(errno.toString())
            else -> bytesRead
        }
    }

    override fun close() {
        if (closed) return
        closed = true
        fclose(file)
    }
}

@OptIn(ExperimentalForeignApi::class, UnsafeIoApi::class)
internal class FopenZeroCopyFileSink(path: Path) : RawSink {
    private val file = fopen(path.toString(), "wb")!!
    private var closed = false

    @OptIn(UnsafeIoApi::class)
    override fun write(
        source: Buffer,
        byteCount: Long
    ) {
        require(byteCount >= 0L) { "byteCount < 0: $byteCount" }
        require(source.size >= byteCount) { "source.size=${source.size} < byteCount=$byteCount" }
        check(!closed) { "closed" }

        var remaining = byteCount
        while (remaining > 0) {
            remaining -= UnsafeBufferOperations.readFromHead(source) { bytes, startIndex, endIndex ->
                val toWrite = minOf((endIndex - startIndex).toLong(), remaining)
                val bytesWritten = bytes.usePinned {
                    fwrite(it.addressOf(startIndex), 1u, toWrite.convert(),file).toInt()
                }
                if (bytesWritten == -1) {
                    throw IOException(errno.toString())
                }
                bytesWritten
            }
        }
    }

    @OptIn(ExperimentalForeignApi::class)
    override fun flush() {
        if (fflush(file) != 0) {
            throw IOException(errno.toString())
        }
    }

    override fun close() {
        if (closed) return
        closed = true
        if (fclose(file) != 0) {
            throw IOException(errno.toString())
        }
    }
}
~~~

</details>

Please tell me what do you think?

**Comment by MohammedKHC** [comment]
@fzhinkin Any updates?

**Comment by eyupcanakman** [comment]
The flat read numbers have a mechanical explanation.
`glibc` skips its internal buffer whenever an `fread` request is at least the buffer size and reads straight into the destination (`_IO_file_xsgetn` in `libio/fileops.c`).
Segment-sized reads were already going direct, so the read path had no copy to remove.

Here is an `fread` loop with 8 KiB requests under `strace` on `glibc` 2.36, every read lands in the caller's buffer.

```
USERBUF 0xc43db7ac1480..0xc43db7ac3480 (chunk=8192)
read(0x3, 0xc43db7ac1480, 0x2000) = 0x2000
read(0x3, 0xc43db7ac1480, 0x2000) = 0x2000
```

Writes are different.
An 8 KiB `fwrite` against the default 4 KiB buffer becomes a `memcpy` into the `stdio` buffer plus two `write` calls, which matches the write improvement you measured.
So your benchmark is fine, the win is write-side only.
The read-side argument for the change is fd-level control, things like `O_CLOEXEC` (#493) and tty behavior.

fzhinkin (approved):

fzhinkin:
Sorry for delays, it all looks good.

PR #508, opened 2026-05-20
Drop workaround for Android 4.2.2 getsockname bug

MohammedKHC:
There is pretty much no reason to support Android <= 4.2.2 nowadays.


fzhinkin:
Sorry for the delay and thanks for the contribution!

We claim to support Android 5+:
> ... the library is compatible with Android 5.0+ (API level 21+).

So it should be fine to remove the workaround for an unsupported version.

fzhinkin (approved):

PR #509, opened 2026-05-21
Use bswap intrinsics for reverseBytes() on native targets to better utilize hardware

KitsuneAlex:
Kotlin/Native has support for GCC/Clang builtins, including the various `bswap` intrinsics which may compile down to `BSWAPnn` instructions on x86 directly (and `REVnn` on aarch64).
These intrinsics should be used to implement the `reverseByte()` extensions on Native targets to get the maximum hardware support possible, since common implementations may either be slower or can't be optimized that well.

### DIFF
--- a/core/native/src/UtilsNative.kt
+++ b/core/native/src/UtilsNative.kt
@@ -5,6 +5,10 @@
 
 package kotlinx.io
 
-internal actual fun Short.reverseBytes(): Short = reverseBytesCommon()
-internal actual fun Int.reverseBytes(): Int = reverseBytesCommon()
-internal actual fun Long.reverseBytes(): Long = reverseBytesCommon()
+import platform.builtin.builtin_bswap16
+import platform.builtin.builtin_bswap32
+import platform.builtin.builtin_bswap64
+
+internal actual fun Short.reverseBytes(): Short = builtin_bswap16(this)
+internal actual fun Int.reverseBytes(): Int = builtin_bswap32(this)
+internal actual fun Long.reverseBytes(): Long = builtin_bswap64(this)


fzhinkin:
Are there any actual performance benefits of this change? For release binaries, LLVM recognizes the byte swap pattern and replaces it with the optimal implementation (for example, `swpal` on arm64).

KitsuneAlex:
I don't know why you would ever rely on the compiler recognizing patterns in this case, when you already have expect-actuals which are being used for the JVM, but not for the native version. Why gamble on optimizations when you don't have to? ^^

fzhinkin:
Why complicating the implementation (even in a slightest), if it does not affect the resulting code? ;)
And yeah, these functions should be in the stdlib, I know.

Anyways, it's worth checking that the intrinsic calls how the same or better performance compared to the existing implementation.

fzhinkin:
> Why complicating the implementation (even in a slightest)

The change does not make the implementation less trivial, compared to what it was, though.

So it's mostly the matter of its performance characteristics.

KitsuneAlex:
I definitly do not agree with you there. Using a single intrinsic builtin, instead of using a handrolled, error prone version of the same function (even if it is tested already) is definitly more trivial. Both in terms of cognitive complexity and in terms of LoC.

Apart from the fact, that assuming LLVM does optimize exactly to the same code, the worst case would be the performance being the same as before. The best case is we get some marginal improvements depending on the system configuration.

Another reason why using intrinsics is good, even if not relevant to most real world cases, is compile time.
Instead of letting the very long compilation pipeline figure out the pattern of the handrolled version, just be more precise right off the bat.

Of course i can write some tests/benchmarks if you like before deciding on how to procede with this change.

KitsuneAlex:
Also, you already have the expect/actuals in place, and you optimized them for JVM, why not do the same for native? You could argue the JVM can optimize the common code just fine too.

BeastLe9enD:
I mean whats the point on not doing it? If it would be a huge change, I could see the point. But with +7 -3 lines its really obvious whats happening here. So why hope that LLVM optimizes in some release mode when its a direct ReverseBytes -> bswap Instruction?

fzhinkin:
> Using a single intrinsic builtin, instead of using a handrolled, error prone version of the same function (even if it is tested already) is definitly more trivial. Both in terms of cognitive complexity and in terms of LoC.

If you believe there are some correctness problems with the existing implementation, it's worth fixing it as it we continue using it for other targets.

> Apart from the fact, that assuming LLVM does optimize exactly to the same code, the worst case would be the performance being the same as before. The best case is we get some marginal improvements depending on the system configuration.

Please check it, if performance remains the same or improves, I don't mind merging the PR.

> You could argue the JVM can optimize the common code just fine too.

JVM does not recognize bswap patterns, that's the difference.

KitsuneAlex:
I will write some benchmarks ^^

KitsuneAlex:
I did some benchmarks for this exact thing in my bitstream processing library, and i have to apologize officially.
I was right with the assumption that the intrinsic may aid the generation of more optimized code in _some_ cases, but since intrinsics or code in general can't be inlined beyond the CInterop boundary, the interop overhead destroys all the benefits.

KitsuneAlex:
One interesting thing i found out tho: byte swapping seems to be more efficient in Kotlin code no matter the data type, however, if you want to perform bit reversal on a 32 or 64 bit type, using the LLVM builtins yield around 25-30% improvement on my current setup. I'll publish the benchmarks along the new update of the library i am working and i'll link it here if you are curious.

fzhinkin:
@KitsuneAlex thank you for investing it.

> I'll publish the benchmarks along the new update of the library i am working and i'll link it here if you are curious.

Absolutely! Please share your findings.

Also, if you have use cases for both reversal and byte-swap, I'd appreciate if you can share it here (or in https://youtrack.jetbrains.com/issue/KT-51721) - these functions were candidates for inclusion into the standard library, but we never had a critical number of users for them :/

Issue #301, opened 2024-04-25
Map each code point of an ill-formed UTF-8 subsequence to a replacement character individually


fzhinkin:
As it was pointed out in https://github.com/Kotlin/kotlinx-io/pull/290#discussion_r1567268068, kotlinx-io converts different ill-formed UTF-8 subsequences differently: either the whole multi-code-point subsequence replaced with a single replacement character, or each code points is converted separately:
- `0xf0 0x89 0x89 <EOF>` -> `�`
-  `0xf0 0x89 0x89 0x89 <EOF>` -> `�`
- `0xf0 0xf0 0xf0 <EOF>` -> `���`

The UTF-8 spec allows handling these ill-formed sequences whatever way we want as long as errors are somehow reported. However, such behavior looks a bit inconsistent and it's hard to reason about how an arbitrary byte sequences will be converted.

We should improve the way ill-formed sequences are handled and stick to an approach adopted by other languages/libraries: convert only ill-formed subsequences consisting of a single byte.

That's how it's done in:
- Java:
```
jshell> new String(new byte[]{(byte)0xf0,(byte)0x89,(byte)0x89,(byte)0x89})
$5 ==> "����"
```
- Python 3:
```
>>> b'\xf0\x89\x89\x89'.decode("utf-8", errors='replace')
'����
```
- Go: 
```
fmt.Println(string([]byte{0xf0, 0x89, 0x89, 0x89}))
...

����
```

ilya-g:
See also the recommendation "U+FFFD Substitution of Maximal Subparts" in https://www.unicode.org/versions/Unicode15.1.0/ch03.pdf

fzhinkin:
It seems like `kotlinx-io` behavior could be aligned w/ Kotlin Stdlib (`ByteArray.decodeToString` in particular) in all scenarios except surrogate code points handling. On JVM, byte-sequences encoding surrogate code points are replaced with a single `�`, on all other platforms with `���`:
```
ubyteArrayOf(0xedu, 0xbfu, 0xbfu).asByteArray().decodeToString()
```

https://pl.kotl.in/LMjgmMVGX

PR #516, opened 2026-07-20
intrinsic accelerated ASCII dominant content

### LINKED ISSUE #515: Source.readString() bypasses JVM String decoding intrinsics; readString(charset) is 3-4x faster for UTF-8
## Summary

The common no-arg `Source.readString()` decodes UTF-8 through `commonToUtf8String`: it stages a `CharArray`, decodes byte by byte in Kotlin, and the `String(chars)` constructor then compresses the chars back to Latin-1 for ASCII content (compact strings). Three passes over the data.

On the JVM, the platform overload `readString(charset)` already takes a much better route: `String(bytes, charset)`, which for UTF-8 ASCII is an intrinsic-accelerated scan (`hasNegatives`) plus an array copy. For typical JSON payloads it is 3.3-4.5x faster and allocates about half as much.

Would you be open to specializing the UTF-8 path of the no-arg `readString()` on the JVM (expect/actual or an internal hook) so it uses the same construction as `readString(charset)`? Happy to contribute the PR if so.

## Benchmarks

JMH 1.37, avgt, 2 forks, JDK 21, arm64 macOS, kotlinx-io 0.9.1. Payloads are realistic ASCII JSON (a 30-field object, repeated with per-element variation for the larger sizes); "mixed" additionally contains 2-3 byte chars and an emoji. Source-based variants include identical Buffer staging cost.

| Payload | `readString()` (common) | `readString(UTF_8)` (JVM overload) | `String(bytes, UTF_8)` (floor, no Source) |
|---|---|---|---|
| 960 B | 219 ns | 50 ns | 27 ns |
| 16 KB | 4,369 ns | 1,322 ns | — |
| 100 KB | 25,665 ns | 7,766 ns | 2,281 ns |
| 16 KB mixed | 5,293 ns | 5,686 ns | — |

Allocation (gc.alloc.rate.norm):

| Payload | `readString()` | `readString(UTF_8)` |
|---|---|---|
| 960 B | 2,936 B/op | 1,032 B/op |
| 16 KB | 65,472 B/op | 32,776 B/op |
| 100 KB | 402,592 B/op | 201,336 B/op |

Two honest caveats from the data:

- For non-ASCII-heavy content the two paths are at parity (5.3 vs 5.7 µs on the mixed payload), so the win is specifically ASCII-dominant content, which JSON, HTML, and most protocol text are.
- Malformed-input replacement conventions differ in one corner: for an overlong encoding (e.g. `0xC0 0xAF`), the common decoder emits one U+FFFD while the JDK emits two. All other cases I checked (truncated 2/3/4-byte sequences, lone continuation bytes, CESU-8 surrogate bytes, stray continuation runs) produce identical output. So the switch is observable for that malformed corner case.

## Repro

```kotlin
private val bytes = /* ~100 KB of ASCII JSON */
private fun srcOf(b: ByteArray): Buffer = Buffer().apply { write(b) }

@Benchmark fun common(): String = srcOf(bytes).readString()
@Benchmark fun jvmOverload(): String = srcOf(bytes).readString(Charsets.UTF_8)
@Benchmark fun floor(): String = String(bytes, Charsets.UTF_8)
```

## Context

Found while fixing the encode-side analog in Ktor ([KTOR-9704](https://youtrack.jetbrains.com/issue/KTOR-9704), PR [ktorio/ktor#5743](https://github.com/ktorio/ktor/pull/5743): `String.toByteArray(UTF_8)` used the validating Kotlin encode loop instead of the `String.getBytes` intrinsic, 44-87x on ASCII). Ktor's `readText(UTF_8)` delegates to `readString()`, so ktor inherits the decode gap too; we can route ktor to the JVM overload ourselves, but fixing it here would benefit every kotlinx-io consumer.

inemtsev:
**Motivation**

The common no-arg Source.readString() decodes UTF-8 through commonToUtf8String, staging a CharArray and decoding byte by byte. On modern JVMs, String(bytes, offset, length, UTF_8) is intrinsic-accelerated (compact strings): ASCII-dominant content (typical JSON, HTML, protocol text) decodes in bulk, 3-4.5x faster with about half the allocations.

**Solution**

Per the discussion in #515, the UTF-8 decode path is specialized through the multi-release jar mechanism, reusing the jvm9Main/jvm9Test compilation setup from the array-exts branch:

- commonReadUtf8 now decodes through an internal expect fun decodeUtf8; all non-JVM actuals delegate to the existing common decoder unchanged.
- The JVM actual routes through Utf8Intrinsics: the base variant (JDK 8, Android) delegates to the common decoder, while a jvm/src9 variant packed into META-INF/versions/9 uses String(bytes, offset, length, UTF_8).
- No public API changes; ABI dumps are untouched.

Behavior note: on JDK 9+, ill-formed UTF-8 sequences are substituted following the JDK decoder's Unicode "maximal subpart" convention, which for some inputs (e.g. the overlong encoding 0xC0 0xAF) emits a different number of U+FFFD than the common decoder (two instead of one). This is the convention requested in #301. All well-formed input decodes identically; JDK 8 and non-JVM targets are unchanged.

**Benchmarks**
(the table above, plus:) JDK 8 stays on the base path and is unchanged within noise. Non-ASCII-heavy content ("mixed") pays 6–10% on JDK 9+; ASCII-dominant payloads gain 2–4.9x from 128 B up. An earlier exploration also confirmed JDK 9 itself never regresses (Zulu 9.0.7 under Rosetta with a matched control: 1.1–2.1x wins), and the no-arg path now slightly outperforms the JVM readString(charset) overload.

| Payload | JDK 8 | JDK 11 | JDK 21 | JDK 26 |
|---|---|---|---|---|
| 16 B | 15.7 → 15.8 (1.0x) | 16.5 → 14.5 (1.14x) | 15.8 → 12.5 (1.26x) | 15.4 → 13.8 (1.12x) |
| 128 B | 40.7 → 40.4 (1.0x) | 41.8 → 20.8 (2.0x) | 40.6 → 18.2 (2.2x) | 40.4 → 19.7 (2.1x) |
| 960 B | 234.6 → 235.7 (1.0x) | 213.6 → 48.0 (4.5x) | 224.9 → 45.6 (4.9x) | 222.1 → 46.5 (4.8x) |
| 16 KB | 4,651 → 4,686 (1.0x) | 4,215 → 1,411 (3.0x) | 4,561 → 1,401 (3.3x) | 4,586 → 1,433 (3.2x) |
| 100 KB | 26,284 → 27,442 (1.0x) | 26,742 → 8,283 (3.2x) | 26,520 → 8,520 (3.1x) | 26,694 → 8,528 (3.1x) |
| 16 KB mixed | 5,153 → 5,190 (1.0x) | 5,240 → 5,805 (0.90x) | 5,331 → 5,646 (0.94x) | 5,326 → 5,743 (0.93x) |

It seems that when emoji is included (mixed) there is a small penalty. Probably because this forces the result out of compact Latin-1 storage into UTF-16. I dug deeper to see how much of the string needs to be non-ASCII to hit the bad-case. Results:

| Non-ASCII density | JDK 11 | JDK 17 | JDK 21 |
|---|---|---|---|
| 0% (pure ASCII) | 4,195 → 1,401 (3.0x) | — | 4,545 → 1,391 (3.3x) |
| 2% Latin | 4,672 → 16,860 (**0.28x**) | 4,475 → 16,634 (**0.27x**) | 4,843 → 4,863 (1.00x) |
| 10% Latin | 6,855 → 15,709 (**0.44x**) | 6,932 → 15,684 (**0.44x**) | 7,183 → 10,841 (0.66x) |
| 33% Latin | 11,499 → 13,150 (0.87x) | — | 11,822 → 9,717 (1.22x) |
| 100% Latin | 5,839 → 5,376 (1.09x) | — | 5,683 → 4,353 (1.31x) |
| 2% CJK | 4,794 → 5,285 (0.91x) | — | 5,128 → 5,220 (0.98x) |
| 10% CJK | 6,797 → 6,603 (1.03x) | — | 7,149 → 7,286 (0.98x) |
| 33% CJK | 11,195 → 9,202 (1.22x) | — | 11,262 → 10,800 (1.04x) |
| 100% CJK | 8,067 → 9,496 (0.85x) | — | 7,343 → 8,357 (0.88x) |

Looks like sparse non-ASCII characters pay a performance penalty if below JVM 21. 

I am not sure, if I should **gate at JVM 21+** or only do this path if **pure ASCII:**

| | `versions/21` gate | hybrid ASCII probe (stays `versions/9`) |
|---|---|---|
| ASCII on JDK 11/17 | 1.0x — win forfeited | ~2.5–3x |
| ASCII on JDK 21+ | 3.3x | ~2.8–3x (probe overhead) |
| Sparse Latin (European text) | 0.66–1.0x on 21+ | 1.0x by construction |
| CJK / emoji | 0.88–1.04x on 21+ | 1.0x by construction |
| Dense Latin (100%) | 1.31x on 21+ | 1.0x — win forfeited |]
| Sensitivity to JDK decoder quality | still exposed (JDK 22+ could shift again) | none — JDK path only ever sees ASCII |
| Code | trivial (dir rename + toolchain bump) | +probe & its tests (~15 lines) |

Hybrid would look like this:

```kotlin
fun decodeUtf8ToString(source: ByteArray, beginIndex: Int, endIndex: Int): String =
    if (containsNonAscii(source, beginIndex, endIndex))
        source.commonToUtf8String(beginIndex, endIndex)      // today's path
    else
        String(source, beginIndex, endIndex - beginIndex, Charsets.ISO_8859_1)
```

@fzhinkin which option do you think is best?
Regards.

Resolves #515

### DIFF
--- a/benchmarks/src/jvmMain/kotlin/ReadStringUtf8GateBenchmark.kt
+++ b/benchmarks/src/jvmMain/kotlin/ReadStringUtf8GateBenchmark.kt
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.io.benchmarks
+
+import kotlinx.benchmark.Benchmark
+import kotlinx.benchmark.Param
+import kotlinx.benchmark.Scope
+import kotlinx.benchmark.Setup
+import kotlinx.benchmark.State
+import kotlinx.io.Buffer
+import kotlinx.io.readString
+
+/**
+ * Compares the common readString() UTF-8 decoder against the JVM readString(charset) overload
+ * across payload sizes, to pick the JDK version gate for the multi-release JAR specialization
+ * discussed in https://github.com/Kotlin/kotlinx-io/issues/515.
+ *
+ * Payloads are ASCII JSON sliced to the exact byte count; "-mixed" additionally contains
+ * 2-3 byte characters and an emoji. Both benchmarks include identical Buffer staging cost.
+ */
+@State(Scope.Benchmark)
+open class ReadStringUtf8GateBenchmark {
+    @Param("16", "128", "960", "16384", "102400", "16384-mixed")
+    var payload: String = ""
+
+    private var bytes: ByteArray = ByteArray(0)
+
+    @Setup
+    fun setup() {
+        val size = payload.removeSuffix("-mixed").toInt()
+        bytes = if (payload.endsWith("-mixed")) mixedJson(size) else asciiJson(size)
+    }
+
+    @Benchmark
+    fun common(): String = Buffer().apply { write(bytes) }.readString()
+
+    @Benchmark
+    fun jvmOverload(): String = Buffer().apply { write(bytes) }.readString(Charsets.UTF_8)
+}
+
+private fun jsonObject(id: Int): String = buildString {
+    append("{\"id\":").append(id)
+    for (f in 1..29) {
+        append(",\"field_").append(f).append("\":")
+        when (f % 4) {
+            0 -> append(id * 31 + f)
+            1 -> append("\"value_").append(id).append('_').append(f).append('"')
+            2 -> append((id + f) % 2 == 0)
+            else -> append(id % 100).append('.').append(f)
+        }
+    }
+    append('}')
+}
+
+private fun asciiJson(size: Int): ByteArray {
+    val sb = StringBuilder("[")
+    var id = 0
+    while (sb.length <= size) {
+        if (id > 0) sb.append(',')
+        sb.append(jsonObject(id++))
+    }
+    return sb.substring(0, size).encodeToByteArray()
+}
+
+private fun mixedJson(size: Int): ByteArray {
+    val sb = StringBuilder()
+    var byteLength = 0
+    var id = 0
+    while (byteLength < size) {
+        val chunk = "{\"città\":\"München_$id\",\"note\":\"日本語テキスト 🚀\"," +
+                jsonObject(id).removePrefix("{") + ","
+        sb.append(chunk)
+        byteLength += chunk.encodeToByteArray().size
+        id++
+    }
+    var s = sb.toString()
+    var encoded = s.encodeToByteArray()
+    while (encoded.size > size) {
+        s = s.substring(0, s.length - 1)
+        encoded = s.encodeToByteArray()
+    }
+    // Pad with spaces at the byte level to hit the exact size; the result stays valid UTF-8.
+    val out = encoded.copyOf(size)
+    out.fill(' '.code.toByte(), encoded.size, size)
+    return out
+}
--- a/benchmarks/src/jvmMain/kotlin/Utf8DensitySweepBenchmark.kt
+++ b/benchmarks/src/jvmMain/kotlin/Utf8DensitySweepBenchmark.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.io.benchmarks
+
+import kotlinx.benchmark.Benchmark
+import kotlinx.benchmark.Param
+import kotlinx.benchmark.Scope
+import kotlinx.benchmark.Setup
+import kotlinx.benchmark.State
+import kotlinx.io.Buffer
+import kotlinx.io.readString
+
+/**
+ * Measures readString() over a fixed 16 KiB payload while sweeping the density of
+ * non-ASCII characters, to locate the crossover between the JDK decoder's bulk-ASCII
+ * advantage and its cost on scattered multi-byte content.
+ *
+ * Kinds: "ascii" is a pure-ASCII baseline; "latin-N" replaces N% of characters with
+ * a 2-byte char (U+00E9, stays Latin-1-compact on the JVM); "cjk-N" uses a 3-byte
+ * char (U+65E5, forces UTF-16 storage). Characters are scattered evenly.
+ */
+@State(Scope.Benchmark)
+open class Utf8DensitySweepBenchmark {
+    @Param("ascii", "latin-2", "latin-10", "latin-33", "latin-100", "cjk-2", "cjk-10", "cjk-33", "cjk-100")
+    var kind: String = ""
+
+    private var bytes: ByteArray = ByteArray(0)
+
+    @Setup
+    fun setup() {
+        bytes = sweepPayload(kind, 16384)
+    }
+
+    @Benchmark
+    fun readString(): String = Buffer().apply { write(bytes) }.readString()
+}
+
+private fun sweepPayload(kind: String, size: Int): ByteArray {
+    fun baseChar(i: Int): Char = if (i % 8 == 7) ' ' else ('a' + (i * 7) % 26)
+
+    val everyNth: Int
+    val replacement: Char
+    if (kind == "ascii") {
+        everyNth = 0
+        replacement = ' '
+    } else {
+        val parts = kind.split('-')
+        replacement = if (parts[0] == "latin") 'é' else '日'
+        everyNth = 100 / parts[1].toInt()
+    }
+
+    val sb = StringBuilder()
+    var byteLength = 0
+    var i = 0
+    while (byteLength < size) {
+        val c = if (everyNth > 0 && i % everyNth == 0) replacement else baseChar(i)
+        sb.append(c)
+        byteLength += if (c.code < 0x80) 1 else if (c.code < 0x800) 2 else 3
+        i++
+    }
+    var s = sb.toString()
+    var encoded = s.encodeToByteArray()
+    while (encoded.size > size) {
+        s = s.substring(0, s.length - 1)
+        encoded = s.encodeToByteArray()
+    }
+    // Pad with spaces at the byte level to hit the exact size; the result stays valid UTF-8.
+    val out = encoded.copyOf(size)
+    out.fill(' '.code.toByte(), encoded.size, size)
+    return out
+}
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -7,6 +7,9 @@
 
 import org.gradle.internal.os.OperatingSystem
 import org.jetbrains.kotlin.gradle.ExperimentalWasmDsl
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinJvmCompilation
+import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
 
 plugins {
     id("kotlinx-io-multiplatform")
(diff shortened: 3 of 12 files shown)


fzhinkin:
Thanks for opening the PR. I'll review it the next week.

fzhinkin:
@inemtsev, can you please elaborate on how you measured the performance? A jar file built for `:kotlinx-io-benchmarks` does not enable MRJ, and if benchmarks are executed via Gradle commands, they utilize a JDK version corresponding to a configured toolchain (`8`).

Also, please include a JSON file with benchmarking results (or, at least, full text output). Mean values (those are average time, ns/op, right?) without corresponding CIs are pretty useless and don't show the full picture.

> I dug deeper to see how much of the string needs to be non-ASCII to hit the bad-case. 

Could you please share the corresponding benchmark?

> Hybrid would look like this
> ...

Wouldn't checking the underlying encoding be too heavyweight, especially for longer inputs?

inemtsev:
> @inemtsev, can you please elaborate on how you measured the performance? A jar file built for `:kotlinx-io-benchmarks` does not enable MRJ, and if benchmarks are executed via Gradle commands, they utilize a JDK version corresponding to a configured toolchain (`8`).
> 
> Also, please include a JSON file with benchmarking results (or, at least, full text output). Mean values (those are average time, ns/op, right?) without corresponding CIs are pretty useless and don't show the full picture.
> 
> > I dug deeper to see how much of the string needs to be non-ASCII to hit the bad-case.
> 
> Could you please share the corresponding benchmark?
> 
> > Hybrid would look like this
> > ...
> 
> Wouldn't checking the underlying encoding be too heavyweight, especially for longer inputs?

You are correct, I noticed this in my tests also, so the benchmarks were not run through Gradle. Gradle was used only to build the jars then each benchmark run is a direct java invocation of one specific installed JDK. 

The packaged kotlinx-io-core jar (which is MRJ-enabled, `Multi-Release: true` in its manifest) was prepended on the classpath ahead of the JMH fat jar, so the versioned Utf8Intrinsics wins class resolution; I verified the fast path was actually engaged via the 0xC0 0xAF decode discriminator from ReadStringJdk9Test.

I will post clearer, more comprehensive benchmarks shortly.

inemtsev:
[pr516-benchmarks.zip](https://github.com/user-attachments/files/30628510/pr516-benchmarks.zip)
I redid the benchmarks with proper statistics, with same conclusions, attached. 18 JMH JSON result files (5 forks, 5x1s warmup + 5x1s measurement per fork, 99.9% CIs), the complete text output of every run (each log begins with the exact command line and `java -version`), the runner script, and the decode-path probe source. Every measurement was preceded by the `0xC0 0xAF` probe described in my previous comment, asserting which decoder was live on that exact classpath; the run log records all 18 checks: `before` jars hit the common decoder everywhere, `after` jars hit the JDK decoder on 9+ and the common decoder on JDK 8.

**Gate benchmark**: `ReadStringUtf8GateBenchmark.common`, avg ns/op, `before → after (before/after)`:

| payload | JDK 8 | JDK 11 | JDK 17 | JDK 21 | JDK 25 |
|---|---|---|---|---|---|
| 16 | 15.7 ± 0.1 → 16.0 ± 0.1 (0.98x) | 16.4 ± 0.1 → 14.5 ± 0.1 (1.13x) | 15.5 ± 0.1 → 11.7 ± 0.1 (1.32x) | 15.7 ± 0.3 → 12.5 ± 0.1 (1.26x) | 15.9 ± 0.7 → 12.2 ± 0.2 (1.31x) |
| 128 | 40.9 ± 1.5 → 41.6 ± 1.8 (0.98x) | 41.3 ± 0.2 → 20.4 ± 0.2 (2.02x) | 38.4 ± 0.2 → 17.7 ± 0.1 (2.17x) | 40.1 ± 0.2 → 18.0 ± 0.2 (2.23x) | 40.0 ± 0.2 → 17.8 ± 0.1 (2.25x) |
| 960 | 240.5 ± 8.8 → 243.9 ± 21.5 (0.99x) | 215.5 ± 2.7 → 47.1 ± 0.2 (4.58x) | 205.5 ± 0.8 → 45.6 ± 0.6 (4.51x) | 220.1 ± 0.7 → 45.8 ± 0.7 (4.81x) | 221.2 ± 0.9 → 44.4 ± 0.4 (4.99x) |
| 16384 | 4,794 ± 35 → 4,602 ± 59 (1.04x) | 4,122 ± 37 → 1,361 ± 18 (3.03x) | 4,058 ± 46 → 1,307 ± 7 (3.11x) | 4,378 ± 26 → 1,301 ± 8 (3.37x) | 4,385 ± 26 → 1,302 ± 9 (3.37x) |
| 102400 | 27,403 ± 211 → 26,011 ± 330 (1.05x) | 25,010 ± 721 → 8,201 ± 140 (3.05x) | 24,138 ± 192 → 7,996 ± 173 (3.02x) | 25,927 ± 92 → 7,793 ± 57 (3.33x) | 25,778 ± 115 → 7,860 ± 57 (3.28x) |
| 16384-mixed | 5,358 ± 122 → 5,026 ± 17 (1.07x) | 5,054 ± 25 → 5,633 ± 14 (0.90x) | 5,064 ± 36 → 5,261 ± 126 (0.96x) | 5,067 ± 21 → 5,359 ± 20 (0.95x) | 5,075 ± 12 → 5,468 ± 36 (0.93x) |

**Density sweep**: `Utf8DensitySweepBenchmark.readString`, 16 KiB payload:

| kind | JDK 11 | JDK 17 | JDK 21 | JDK 25 |
|---|---|---|---|---|
| ascii | 4,027 ± 61 → 1,302 ± 15 (3.09x) | 4,053 ± 84 → 1,281 ± 10 (3.16x) | 4,333 ± 26 → 1,273 ± 9 (3.40x) | 4,337 ± 24 → 1,278 ± 10 (3.39x) |
| latin-2 | 4,432 ± 39 → 16,463 ± 63 (0.27x) | 4,333 ± 27 → 16,395 ± 51 (0.26x) | 4,664 ± 21 → 4,682 ± 24 (1.00x) | 4,718 ± 23 → 4,687 ± 16 (1.01x) |
| latin-10 | 6,711 ± 49 → 15,339 ± 47 (0.44x) | 6,752 ± 26 → 15,303 ± 53 (0.44x) | 7,031 ± 34 → 10,544 ± 21 (0.67x) | 7,046 ± 23 → 10,565 ± 29 (0.67x) |
| latin-33 | 11,255 ± 108 → 12,865 ± 39 (0.87x) | 11,657 ± 85 → 12,900 ± 34 (0.90x) | 11,597 ± 49 → 9,527 ± 10 (1.22x) | 11,574 ± 35 → 9,530 ± 25 (1.21x) |
| latin-100 | 5,637 ± 43 → 5,220 ± 83 (1.08x) | 5,644 ± 36 → 4,437 ± 35 (1.27x) | 5,367 ± 71 → 4,228 ± 29 (1.27x) | 5,457 ± 73 → 4,232 ± 35 (1.29x) |
| cjk-2 | 4,531 ± 24 → 4,855 ± 28 (0.93x) | 4,599 ± 32 → 4,613 ± 59 (1.00x) | 4,876 ± 24 → 4,937 ± 31 (0.99x) | 4,877 ± 28 → 4,932 ± 34 (0.99x) |
| cjk-10 | 6,595 ± 41 → 6,419 ± 16 (1.03x) | 6,700 ± 36 → 6,071 ± 66 (1.10x) | 6,970 ± 20 → 7,034 ± 23 (0.99x) | 6,994 ± 24 → 7,051 ± 28 (0.99x) |
| cjk-33 | 10,902 ± 46 → 9,086 ± 37 (1.20x) | 10,672 ± 59 → 9,262 ± 81 (1.15x) | 10,887 ± 71 → 10,290 ± 33 (1.06x) | 10,848 ± 70 → 10,289 ± 24 (1.05x) |
| cjk-100 | 7,822 ± 65 → 9,244 ± 21 (0.85x) | 7,940 ± 87 → 8,737 ± 55 (0.91x) | 7,006 ± 28 → 7,951 ± 37 (0.88x) | 6,980 ± 24 → 7,924 ± 12 (0.88x) |

inemtsev:
Here is the investigation of impact of checking encoding ahead of time.

| row | JDK 11 | JDK 21 |
|---|---|---|
| gate 16 B | +25.3% | +19.8% |
| gate 128 B | +15.4% | +19.0% |
| gate 960 B | +31.3% | +9.2% |
| gate 16 KiB | +0.3% | +1.6% |
| gate 100 KiB | +4.0% | +5.4% |
| sweep ascii 16 KiB | +4.2% | +3.2% |

So for longer inputs, where your concern was, the scan costs 0.3-5.4%. It is proportionally most visible at small payloads, where it is a few nanoseconds in absolute terms (e.g. 47.1 → 61.8 ns at 960 B on JDK 11) and the hybrid still keeps a 1.75-4.4x win over the current decoder.

Implementation of encoding check:

```kotlin
internal object Utf8Intrinsics {
    fun decodeUtf8ToString(source: ByteArray, beginIndex: Int, endIndex: Int): String =
        if (containsNonAscii(source, beginIndex, endIndex)) {
            source.commonToUtf8String(beginIndex, endIndex)
        } else {
            String(source, beginIndex, endIndex - beginIndex, Charsets.ISO_8859_1)
        }

    // Every byte's high bit: a non-zero AND against an 8-byte word means non-ASCII.
    private const val HIGH_BITS = -0x7f7f7f7f7f7f7f80L // 0x8080808080808080

    private fun containsNonAscii(source: ByteArray, beginIndex: Int, endIndex: Int): Boolean {
        var i = beginIndex
        if (endIndex - i >= 8) {
            val view = ByteBuffer.wrap(source).order(ByteOrder.nativeOrder())
            val swarLimit = endIndex - 8
            while (i <= swarLimit) {
                if (view.getLong(i) and HIGH_BITS != 0L) return true
                i += 8
            }
        }
        while (i < endIndex) {
            if (source[i] < 0) return true
            i++
        }
        return false
    }
}
```

What the probe buys (ratios vs the current decoder, most affected inputs; "latin-tail" is the probe's worst case, 16 KiB of ASCII with a single 2-byte char as the final bytes, forcing a full scan and then a full fallback):

| input | JDK 11 after | JDK 11 hybrid | JDK 21 after | JDK 21 hybrid |
|---|---|---|---|---|
| pure ascii 16 KiB | 3.09x | 2.97x | 3.40x | 3.30x |
| latin-2 | 0.27x | 0.98x | 1.00x | 0.99x |
| latin-10 | 0.44x | 0.99x | 0.67x | 1.00x |
| latin-tail | 0.24x | 0.96x | 2.47x | 0.95x |
| cjk-100 | 0.85x | 0.99x | 0.88x | 0.99x |
| 16 KiB mixed (gate) | 0.90x | 1.00x | 0.95x | 0.98x |

Ratios are speedups vs the current decoder, so values below 1x are regressions. "after" is this PR as it stands today, always handing the bytes to the JDK decoder. "hybrid" is the variant above: run the encoding check first, then the JDK path for pure ASCII or the existing common decoder for everything else.

@fzhinkin I think it's better to go with the hybrid at versions/9 rather than gating at versions/21: JDK 11/17 users keep the ASCII win and no input class regresses. If you agree I can update the PR to this variant.

PR #517, opened 2026-07-21
Avoid lambda allocation for JVM closed checks

MukjepScarlet:
The old implementation will create method ref (for `Buffer`) and a new class for `RealSink`. New implementation is single simple static method.

Reduced 4 anonymous classes

Old example (decomp with JADX):
```java
    @NotNull
    public static final ReadableByteChannel asByteChannel(@NotNull final Source $this$asByteChannel) throws NoWhenBranchMatchedException {
        Function0 function0;
        Intrinsics.checkNotNullParameter($this$asByteChannel, "<this>");
        if ($this$asByteChannel instanceof buffered) {
            function0 = (Function0) new MutablePropertyReference0Impl($this$asByteChannel) { // from class: kotlinx.io.SourcesJvmKt$asByteChannel$isClosed$1
                public Object get() {
                    return Boolean.valueOf(((buffered) this.receiver).closed);
                }

                public void set(Object value) {
                    ((buffered) this.receiver).closed = ((Boolean) value).booleanValue();
                }
            };
        } else {
            if (!($this$asByteChannel instanceof Buffer)) {
                throw new NoWhenBranchMatchedException();
            }
            function0 = SourcesJvmKt::asByteChannel$lambda$0;
        }
        final Function0 isClosed = function0;
        return new ReadableByteChannel() { // from class: kotlinx.io.SourcesJvmKt.asByteChannel.1
            @Override // java.nio.channels.Channel, java.io.Closeable, java.lang.AutoCloseable
            public void close() {
                $this$asByteChannel.close();
            }

            @Override // java.nio.channels.Channel
            public boolean isOpen() {
                return !((Boolean) isClosed.invoke()).booleanValue();
            }

            @Override // java.nio.channels.ReadableByteChannel
            public int read(ByteBuffer sink) {
                Intrinsics.checkNotNullParameter(sink, "sink");
                return SourcesJvmKt.readAtMostTo($this$asByteChannel, sink);
            }
        };
    }

    private static final boolean asByteChannel$lambda$0() {
        return false;
    }
```

### DIFF
--- a/core/jvm/src/SinksJvm.kt
+++ b/core/jvm/src/SinksJvm.kt
@@ -56,27 +56,20 @@ public fun Sink.writeString(string: String, charset: Charset, startIndex: Int = 
  */
 @OptIn(DelicateIoApi::class)
 public fun Sink.asOutputStream(): OutputStream {
-    val isClosed: () -> Boolean = when (this) {
-        is RealSink -> this::closed
-        is Buffer -> {
-            { false }
-        }
-    }
-
     return object : OutputStream() {
         override fun write(byte: Int) {
-            if (isClosed()) throw IOException("Underlying sink is closed.")
+            if (this@asOutputStream.isClosed()) throw IOException("Underlying sink is closed.")
             writeToInternalBuffer { it.writeByte(byte.toByte()) }
         }
 
         override fun write(data: ByteArray, offset: Int, byteCount: Int) {
-            if (isClosed()) throw IOException("Underlying sink is closed.")
+            if (this@asOutputStream.isClosed()) throw IOException("Underlying sink is closed.")
             writeToInternalBuffer { it.write(data, offset, offset + byteCount) }
         }
 
         override fun flush() {
             // For backwards compatibility, a flush() on a closed stream is a no-op.
-            if (!isClosed()) {
+            if (!this@asOutputStream.isClosed()) {
                 this@asOutputStream.flush()
             }
         }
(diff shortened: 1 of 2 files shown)


core/jvm/src/SinksJvm.kt
```diff
@@ -56,27 +56,20 @@ public fun Sink.writeString(string: String, charset: Charset, startIndex: Int =
  */
 @OptIn(DelicateIoApi::class)
 public fun Sink.asOutputStream(): OutputStream {
-    val isClosed: () -> Boolean = when (this) {
-        is RealSink -> this::closed
-        is Buffer -> {
-            { false }
-        }
-    }
```
fzhinkin, line 64: Why no simply turn `isClosed` into a function in this exact place?
I mean,
```kotlin
fun Sink.isClosed(): Boolean = when (this) {
   is RealSink -> this.closed
   is Buffer -> false
}
```

fzhinkin:
Nice catch!

core/jvm/src/SinksJvm.kt
MukjepScarlet: This is also OK I think, but the declaration appears 2 times in same file so I extracted them into single one.

JakeWharton: Local functions are implemented as lambdas in the compiler, no? At the very least it would be stateless, so only a single allocation for the process.

MukjepScarlet: Local functions are compiled into static methods, but without `invokedynamic` call
The captured local vars are passed directly to every call of it, for example:

```kotlin
fun ....
    val map = mapOf<Int, Int>()
    fun test() = map.isEmpty()

    val closable = object : AutoCloseable {
        override fun close() {
            test()
        }
    }

```

Result:

```java
            AutoCloseable var48 = new AutoCloseable() {
               public void close() {
                  UtilKt.main$test(map);
               }
            };


   private static final boolean main$test(Map map) {
      return map.isEmpty();
   }
```

(KT 2.4.0, JDK 25)

fzhinkin (approved):

PR #523, opened 2026-09-18
Add Path.div operator fun

### LINKED ISSUE #505: Add Path.div operator fun
Kotlin stdlib has `Path.div(other: String): Path`: https://github.com/JetBrains/kotlin/blob/2.3.20/libraries/stdlib/jdk7/src/kotlin/io/path/PathUtils.kt#L974

It is trivial to add it:
```kotlin
operator fun Path.div(relativePath: String): Path = Path(this, relativePath)
```

ShadowRZ:
Add a convenience function that's present [in stdlib](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io.path/div.html), [and okio](https://lysine.dev/okio/3.x/okio/okio/-path/div.html)

Closes #505


MohammedKHC:
Doesn't it make more sense to make it inline?

ShadowRZ:
> Doesn't it make more sense to make it inline?

IDE gave this warning when appling `inline`:

> Expected performance impact from inlining is insignificant. Inlining works best for functions with parameters of function types.

Also the `kotlinx.io.files.PathsKt#Path(kotlinx.io.files.Path, kotlin.String...)` function also doesn't inline too

